Issue #2, opened 2026-02-20
Play nicer with librustzcash by using Option instead of generic in Bundle


S1nus:
librustzcash [parameterizes bundles with a generic](https://github.com/tachyon-zcash/librustzcash/blob/main/zcash_primitives/src/transaction/mod.rs#L340) for the type of value commitment, but tachyon bundles currently require a second generic for whether or not a stamp is present.

Our [current approach](https://github.com/tachyon-zcash/zebra/pull/25) to aggregation involves putting the aggregated stamp in the _coinbase transaction_, which librustzcash must be able to serialize alongside non-coinbase transactions.

Instead of a generic, we should use `Option<T>`.

PR #4, opened 2026-02-20
use option in bundle

S1nus:
closes https://github.com/tachyon-zcash/tachyon/issues/2

builds on https://github.com/tachyon-zcash/tachyon/pull/1


turbocrime (approved):
rebased to main. approving but i prefer https://github.com/tachyon-zcash/tachyon/pull/12 so that we can maintain type-constrained pathways.

PR #10, opened 2026-02-26
book work

turbocrime:
Previously from the book repo, resurrected into the main repository.

## Book contents

- **Data Model**: bundles, tachygrams
- **Cryptographic Primitives**: key hierarchy, notes, nullifier derivation (GGM tree PRF)
- **Protocol**: authorization (per-action signing, value balance, binding signature), aggregation (autonome/aggregate/adjunct lifecycle)
- **Reference**: domain separators

## Changes in this PR

- Authorization chapter documents the complete signing flow across trust boundaries (user device, custody device, stamper, consensus)
- Alpha derivation uses distinct domain separators for spend (`Tachyon-Spend`) and output (`Tachyon-Output`) actions
- Note commitment renamed from `cmx` to `cm` (Tachyon uses Poseidon hashes, not curve point x-coordinates)
- Domain separator reference table updated with current personalizations
- RCV derivation documented as uniformly random with a note that it may be revised

## Known issues

The authorization chapter documents the current per-action sighash and binding sighash design, which diverges from the Zcash protocol's requirement for a transaction-wide SigHash. See #11 for the tracking issue. The book will need updating when that issue is resolved.


.github/workflows/book.yml
turbocrime: adds mdbook-mermaid, switches to a cargo install action like ragu CI

justfile
turbocrime: adds some justfile tasks for the book like ragu

PR #12, opened 2026-02-26
Flexible bundle types

turbocrime:
Redesigns `Bundle` to use a sealed `StampState` trait instead of a bare generic, keeping compile-time stamp tracking while adding an `Option<Stamp>` escape hatch for mixed contexts.

Previously `Bundle<S, V>` carried two type parameters. This PR:

- Replaces the open generic with a sealed `StampState` trait, restricting `S` to `Stamp`, `Stampless`, or `Option<Stamp>`
- Drops the `V` parameter (value balance is always `i64`)
- Re-introduces `Stamped` and `Stripped` type aliases lost in #4
- Adds `From`/`TryFrom` conversions between typed states and `Bundle<Option<Stamp>>` for ergonomic type erasure and recovery

Additional work from #4 pulled forward:

- Derives `Eq`/`PartialEq` on `Action`, `Signature`, `Proof`, `Stamp`, and `ActionVerificationKey`
- Relaxes `proptest` version to `1` to match librustzcash

## Motivation

PR #4 proposed using `Option<Stamp>` directly, which simplified the API but lost type-level knowledge of stamp disposition. This PR takes the best of both worlds: most code works with concrete `Stamped` or `Stripped` types (preventing misuse at compile time), while code that needs to handle bundles with unknown stamp state can use `Bundle<Option<Stamp>>` and convert back via `TryFrom`.

## Type mapping

| Before | After |
|---|---|
| `Bundle<Stamp, V>` | `Bundle<Stamp>` / `Stamped` |
| `Bundle<Stampless, V>` | `Bundle<Stampless>` / `Stripped` |
| *(no equivalent)* | `Bundle<Option<Stamp>>` |

Supersedes #4.

### DIFF
--- a/crates/tachyon/src/bundle.rs
+++ b/crates/tachyon/src/bundle.rs
@@ -67,7 +67,7 @@ impl From<Stamped> for Bundle<Option<Stamp>> {
 }
 
 impl TryFrom<Bundle<Option<Stamp>>> for Stripped {
-    type Error = &'static str;
+    type Error = Stamped;
 
     fn try_from(bundle: Bundle<Option<Stamp>>) -> Result<Self, Self::Error> {
         match bundle.stamp {
@@ -79,7 +79,14 @@ impl TryFrom<Bundle<Option<Stamp>>> for Stripped {
                     stamp: Stampless,
                 })
             },
-            | Some(_) => Err("stamp is present"),
+            | Some(stamp) => {
+                Err(Stamped {
+                    actions: bundle.actions,
+                    value_balance: bundle.value_balance,
+                    binding_sig: bundle.binding_sig,
+                    stamp,
+                })
+            },
         }
     }
 }
@@ -99,7 +106,7 @@ impl From<Stripped> for Bundle<Option<Stamp>> {
 }
 
 impl TryFrom<Bundle<Option<Stamp>>> for Stamped {
-    type Error = &'static str;
+    type Error = Stripped;
 
     fn try_from(bundle: Bundle<Option<Stamp>>) -> Result<Self, Self::Error> {
         match bundle.stamp {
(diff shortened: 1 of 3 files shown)


crates/tachyon/src/bundle.rs
```diff
@@ -17,15 +21,29 @@ use crate::{
...
+impl sealed::Sealed for Stampless {}
+impl sealed::Sealed for Option<Stamp> {}
+
+impl StampState for Stamp {}
+impl StampState for Stampless {}
+impl StampState for Option<Stamp> {}
+
+/// A Tachyon transaction bundle parameterized by stamp state `S`.
 #[derive(Clone, Debug)]
-pub struct Bundle<S, V> {
```
turbocrime: also removes the value parameter (i thought this might be useful in the past, but now i don't think so).

crates/tachyon/src/bundle.rs
```diff
@@ -1,7 +1,11 @@
 //! Tachyon transaction bundles.
 //!
-//! A bundle is parameterized by its stamp state. Actions are constant through
-//! state transitions; only the stamp is stripped or merged.
+//! A bundle is parameterized by stamp state `S: StampState`.
+//! Actions are constant through state transitions; only the stamp changes.
+//!
+//! - [`Stamped`] — self-contained bundle with a stamp
+//! - [`Stripped`] — stamp removed, depends on an aggregate
+//! - `Bundle<Option<Stamp>>` — erased stamp state for mixed contexts
```
TalDerei: https://github.com/tachyon-zcash/zebra/pull/26/changes#diff-225fef395c927fa1cc5e9e243cfa55ba0bef888918aced5992d9eceac7f98e5eR9-R11 will need to be updated so zebra doesn't have to roll its own `Option<Stamp>` field.

crates/tachyon/src/bundle.rs
TalDerei: `Bundle<Stamp, Amount>` doesn't make sense to me

PR #14, opened 2026-02-27
serialization with same approach as Orchard

S1nus:


S1nus:
closing in favor of https://github.com/tachyon-zcash/tachyon/pull/30

PR #16, opened 2026-03-04
split refactors

turbocrime:
some preliminary work extracted from https://github.com/tachyon-zcash/tachyon/pull/13

### DIFF
--- a/book/src/tachygrams.md
+++ b/book/src/tachygrams.md
@@ -45,7 +45,7 @@ The proof's public output `StampDigest` may carry five field elements:
 
 | Field | Type | Elements | Description |
 | ----- | ---- | -------- | ----------- |
-| `actions_acc` | Pallas point | 2 (x,y) | Pedersen hash over action digests |
+| `action_acc` | Pallas point | 2 (x,y) | Pedersen hash over action digests |
 | `tachygram_acc` | Pallas point | 2 (x,y) | Pedersen hash over tachygrams |
 | `anchor` | $\mathbb{F}_p$ scalar | 1 | Accumulator state reference |
 
@@ -82,8 +82,8 @@ tachygrams $tg_i$, the anchor, and the proof bytes.
 2. **No duplicate tachygrams**: check the tachygram list for repeats
 3. **Action sigs**: verify each $sig_i$ against $rk_i$ (RedPallas)
 4. **Binding sig**: verify against $\sum cv_i$
-5. **Reconstruct header**: build `StampDigest { anchor, actions_acc, tachygram_acc }`
-   - **Recompute actions_acc**: $\sum[\text{Poseidon}(rk_i \| cv_i)] \cdot G_\text{acc}$ from visible actions
+5. **Reconstruct header**: build `StampDigest { anchor, action_acc, tachygram_acc }`
+   - **Recompute action_acc**: $\sum[\text{Poseidon}(rk_i \| cv_i)] \cdot G_\text{acc}$ from visible actions
    - **Recompute tachygram_acc**: $\sum[\text{Poseidon}(tg_i)] \cdot G_\text{acc}$ from listed tachygrams
 6. **Verify proof**: call Ragu `verify(Pcd { proof, data: header })`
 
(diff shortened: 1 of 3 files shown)


crates/tachyon/src/custody.rs
TalDerei: ACK – don't think Orchard has a notion of an custody abstraction (which is why I assume you removed this because recall signing is part of a larger transaction auth flow).

crates/tachyon/src/entropy.rs
```diff
@@ -0,0 +1,164 @@
...
+/// signs with $\mathsf{rsk} = \mathsf{ask} + \alpha$, and a separate
+/// (possibly untrusted) device constructs the proof later using $\theta$
+/// and $\mathsf{cm}$ to recover $\alpha$
+/// ("Tachyaction at a Distance", Bowe 2025).
+#[derive(Clone, Copy, Debug)]
+#[expect(
+    clippy::module_name_repetitions,
+    reason = "ActionEntropy is the established protocol name"
+)]
+pub struct ActionEntropy([u8; 32]);
```
TalDerei: ACK – theta/alpha are action-level entropy

crates/tachyon/src/action.rs
```diff
@@ -117,17 +117,17 @@ impl Action {
         // 3. Alpha (user device derives for proof witness)
         let alpha = theta.spend_randomizer(&cm);
 
-        // 4. Spend authorization (custody derives alpha independently)
-        let (rk, sig) = custody.authorize_spend(cv, theta, &cm, rng)?;
+        // 4. Spend authorization
+        let (rk, sig) = alpha.authorize(ask, cv, rng);
```
TalDerei: ACK – we're currently naively performing signature during action construction (action paths signs before the full transaction exists, and all signatures will eventually sign the entire transaction, sig hash, in https://github.com/tachyon-zcash/tachyon/pull/13).

crates/tachyon/src/note.rs
```diff
@@ -50,6 +51,13 @@ use crate::{
 #[derive(Clone, Copy, Debug)]
 pub struct NullifierTrapdoor(Fp);
 
+impl NullifierTrapdoor {
+    /// Generate a fresh random trapdoor.
+    pub fn random<R: RngCore + CryptoRng>(rng: &mut R) -> Self {
+        Self(Fp::random(rng))
+    }
+}
```
TalDerei: fine for being random now, but can we add a TODO to the effect of `nf = nk.derive(psi, flavor)` where  where flavor will come from the epoch/anchor?

crates/tachyon/src/value.rs
```diff
@@ -65,16 +53,46 @@ impl CommitmentTrapdoor {
         Self(Fq::random(rng))
     }
 
+    /// Commit to spend a value with this trapdoor.
+    ///
+    /// $$\mathsf{cv} = [v]\,\mathcal{V} + [\mathsf{rcv}]\,\mathcal{R}$$
+    ///
+    /// Positive $v$ for spends (balance contributed), negative for
+    /// outputs (balance exhausted).
```
TalDerei:
```suggestion
  /// Positive $v$ for spends (balance contributed).

File: crates/tachyon/src/action.rs [hunk 5]
```diff
@@ -95,17 +95,17 @@ impl Action {
...
+    ///    [`ActionEntropy::spend_randomizer`](entropy::ActionEntropy::spend_randomizer)
     /// 4. Spend authorization:
-    ///    [`Custody::authorize_spend`](crate::custody::Custody::authorize_spend)
+    ///    [`SpendRandomizer::authorize`](crate::entropy::SpendRandomizer::authorize)
     /// 5. Assembly: `Action { cv, rk, sig }` +
     ///    [`ActionPrivate`](crate::witness::ActionPrivate)
-    pub fn spend<R: RngCore + CryptoRng, C: Custody>(
-        custody: &C,
+    pub fn spend<R: RngCore + CryptoRng>(
+        ask: &private::SpendAuthorizingKey,
```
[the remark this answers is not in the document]
turbocrime: it's already understood this branch has several shortcomings, and it only exists to extract refactors that could be conceptually separated from #13. i'd like to avoid adding todos about work that's technically already happened

PR #25, opened 2026-03-04
serde in tachyon

S1nus:
this approach to serialization for tachyon types approximates the idealized separation of concerns between orchard and zebra, assuming no tech debt, see this issue: https://github.com/ZcashFoundation/zebra/issues/8155


turbocrime (commented):
do we need to convert errors to arbitrary strings? i would prefer to use the string representation of the underlying error, if we must use a string.

i also wonder if we should drop the existing from/into bytes impls on all the newtypes, and have serde unwrap the newtype and then directly use the inner type's from/into bytes impls.

otherwise, this is pretty much exactly how i imagined it. would like to hold off merging until i can land the binding signature refactor, and then i'll do any necessary rebase work.

PR #26, opened 2026-03-04
Mock Ragu

turbocrime:
Before #27

Ragu can't be integrated yet due to `rand` and `ff` dependency conflicts. This PR adds a standalone mock crate.


TalDerei (approved):
ACK since this is self-contained.

PR #29, opened 2026-03-05
Reorganize keys

turbocrime:
Stacks on #45

Replaces the flat `keys/private` / `keys/public` module split with a role-based organization that mirrors the Tachyon trust model — who holds the key determines which module it lives in.

| Module | Role | Keys |
| --- | --- | --- |
| `keys/custody` | Restricted secrets held by the custody device | `SpendingKey`, `SpendAuthorizingKey`, `NullifierKey` |
| `keys/planner` | Wallet-side keys that cannot spend | `SpendValidatingKey`, `ActionSigningKey`, `BindingSigningKey`, `PaymentKey`, `NoteMasterKey` |
| `keys/delegate` | Epoch-restricted keys issued to proving services | `ProofDelegateKey`, `ProofAuthorizingKey` |

## Other changes

- Absorb GGM tree logic into `keys/note.rs`; replace `Prefixed` depth marker with `Prefix` / `Leaf` / `Master`
- Move `reddsa` module from `pub` to `pub(crate)`; consumers use the re-exported key types instead
- Tighten `lib.rs` public API surface with explicit `pub use` items and clippy `pub_use` allows per-item instead of crate-wide


TalDerei:
what's the state of this work?

PR #36, opened 2026-03-10
chore: ci, docs, lints, and no_std

turbocrime:
Workflows mostly match ragu.

Tests run on `i686-unknown-linux-gnu` and `thumbv7em-none-eabihf` to catch
platform edge cases.

Rustdoc katex fixed, new doc lints added.

PR #41, opened 2026-03-12
no_std support

S1nus:
Adds no_std support to tachyon, to match orchard and to enable builds on wasm and embedded platforms

- uses `rand_core` instead of `rand`
- create std feature flag, and connect it to the std feature flags of blake2b_simd, rand_core, reddsa
- use `lazy_static!` instead of `LazyLock` in some places


TalDerei (approved):
cross referenced against zebra deps, looks good. except there's no CI validation for no_std (maybe we should fold into https://github.com/tachyon-zcash/tachyon/pull/36), so support could silently break. cc @turbocrime

PR #46, opened 2026-03-24
Proof tree

turbocrime:
Large changeset, sorry. Replaces the previous simplistic `ActionStep` + `MergeStep` with a more realistic tree.

## PCD Steps

0. **DelegationSeed**: () → DelegationMasterHeader
1. **DelegationMasterStep**: (DelegationMasterHeader) → DelegationHeader
2. **OutputStamp**: () → StampHeader
3. **DelegationStep**: (DelegationHeader) → DelegationHeader
4. **NullifierStep**: (DelegationHeader) → NullifierHeader
5. **SpendBind**: (NullifierHeader, NullifierHeader) → SpendHeader
6. **SpendableInit**: (NullifierHeader) → SpendableHeader
7. **SpendableRollover**: (NullifierHeader, NullifierHeader) → SpendableRolloverHeader
8. **SpendableLift**: (SpendableHeader) → SpendableHeader
9. **SpendableEpochLift**: (SpendableHeader, SpendableRolloverHeader) → SpendableHeader
10. **SpendStamp**: (SpendHeader, SpendableHeader) → StampHeader
11. **MergeStamp**: (StampHeader, StampHeader) → StampHeader
12. **StampLift**: (StampHeader) → StampHeader

## What's still stubbed

There are still some shortcuts:

- Pool state is usually a flat polynomial when it should often be a hash chain.
- Anchor still carries a block height field internally. This is at least abstracted.

## What else changed

Bundle serialization changed to use wtxid to refer to aggregates in-block, and generally refactored.

Book has a new `proof-pipeline.md` page. Several existing pages have some updates.

Tests greatly improved. New `test_support` module organizes actor-focused collections of tools.


book/src/proof-pipeline.md
```diff
@@ -0,0 +1,100 @@
...
+
+All anchor-bearing headers embed a shared `Anchor` sub-struct representing pool state at a specific block:
+
+| Field | Type | Description |
+|-------|------|-------------|
+| block_height | BlockHeight | Block height in the pool chain |
+| block_commit | BlockCommit | Per-block tachygram set commitment |
+| pool_commit | PoolCommit | Cumulative epoch tachygram commitment |
+| block_chain | BlockChainHash | `H(prev, block_commit)` every block |
+| epoch_chain | EpochChainHash | `H(prev, pool_commit)` at epoch boundaries |
```
TalDerei: we can seed this with more context; we've discussed anchors being more complex than just a hash mapping to the accumulator state at a certain point. it's a (1) hash of accumulator state at a valid checkpoint, (2) epoch number specifying how far the non-inclusion proofs extend, and (3)  second number specifying how many epochs the stamp is valid for based on how many nullifiers have been revealed. it's also worth documenting that non-inclusion proofs having a built-in expiry tied to this.

book/src/proof-pipeline.md
```diff
@@ -0,0 +1,100 @@
+# Proof Pipeline
+
+Tachyon uses a 17-step PCD (Proof-Carrying Data) pipeline for proof generation and aggregation. Steps are organized into three categories: **seeds** (create proofs from scratch), **transforms** (evolve a single proof), and **fuses** (combine two proofs).
```
TalDerei: still don't have a good intuition for all the different step circuits yet, filed https://github.com/tachyon-zcash/ragu/issues/661.

book/src/proof-pipeline.md
```diff
@@ -0,0 +1,100 @@
...
+| epoch_chain | EpochChainHash | `H(prev, pool_commit)` at epoch boundaries |
+
+## Step Types
+
+### Seeds
+
+| Step | Output | Description |
+|------|--------|-------------|
+| DelegationSeed | DelegationHeader | Initialize GGM tree walk for nullifier delegation |
+| SpendNullifier | SpendNullifierHeader | Derive nullifier pair (epoch E and E+1) via full GGM walk |
```
TalDerei: `SpendNullifier`'s "full" depth GGM walk is probably an example where the witness functions defined here don't cleanly translate to circuits (ie. whether a given step's witness will fit in a single circuit, but I understand the constraint chunking is black-boxed). even without explicit quantitive constraints, we could mark / annotate something that directionally helps inform what would need to be split up.

book/src/proof-pipeline.md
```diff
@@ -0,0 +1,100 @@
...
+SpendableInit(NullifierHeader × PoolHeader)   (cm inclusion + nf non-membership)
+  → SpendableLift(× PoolHeader)               (advance within epoch)
+  → SpendableEpochLift(× SpendableRolloverHeader)  (cross epoch)
+
+SpendNullifier(note, nk, epoch)               (or SpendNullifierFuse)
+  → SpendBind(rcv, α, ak, note, nk)
+  → SpendStamp(× SpendableHeader) → StampHeader
+```
+
+`SpendStamp` enforces `left.nullifiers[0] == right.nf` and `left.epoch == epoch_index(right.anchor.block_height)`, binding the spend to the spendable chain's note and epoch.
```
TalDerei: why does `SpendStamp` only constrain `nullifiers[0]`? what about the second nullifier?

crates/tachyon/src/stamp/header.rs
```diff
@@ -0,0 +1,183 @@
...
+        }
+        if right.block_height.epoch() != left_anchor.block_height.epoch() {
+            return Err(mock_ragu::Error);
+        }
+        if left_anchor.epoch_chain != right.epoch_chain {
+            return Err(mock_ragu::Error);
+        }
+
+        // TODO: block_chain continuity from left to right
+        todo!("block_chain continuity verification");
```
TalDerei: ACK – stubbing continuity checks in `SpendableLift` / `StampLift` / `SpendableEpochLift`.

crates/tachyon/src/stamp/spendable.rs
```diff
@@ -0,0 +1,231 @@
...
+            (),
+        ))
+    }
+}
+
+/// Epoch transition: fuses epoch-final spendable with rollover.
+#[derive(Debug)]
+pub struct SpendableEpochLift;
+
+impl Step for SpendableEpochLift {
```
TalDerei: ACK – anchor's are malleable (like penumbra) and can be updated by appending to the accumulator (before epoch rollover). in principle you could fold `SpendableRollover` + `SpendableEpochLift` into one atomic`SpendableCrossEpoch` step that does rollover + lift in one shot. is there some kind of boundary / race condition violation if this isn't atomic? same observation applies to `SpendableLift` + `StampLift`.

TalDerei (commented):
some initial observations

book/src/proof-pipeline.md
turbocrime: after work this week the anchor is simplified to `(block_height, pool_commit)`

will try to add context also

book/src/proof-pipeline.md
turbocrime: the second nullifier is already proven to derive from the same appropriately scoped key.

and, there's no second nullifier on the right pcd to compare against. this is the last available non-inclusion proof, and it only involves one nullifier.

i do notice that this might not be correct: the proofs will only be available for historical epochs. we might need to compare note_id instead of nf, and then make sure the stamp nullifiers are one more epoch forward.

crates/tachyon/src/stamp/spendable.rs
turbocrime: we've got four main constraints in how we organize these:
- pcd fuse is binary
- computation
- secrets available
- public data available

a `SpendableEpochLift` should mostly happen on the sync service and relies on three distinctly sourced pcd:
- the old `SpendableHeader` (must cheaply init on user with secret keys)
- the nullifier relationship `NullifierHeader` (must complete on sync with scoped keys)
- the new epoch state `PoolHeader` (needs consensus data)

but we can't do a trinary merge. so we stage a `SpendableRolloverHeader` combining the nullifier derivation and the new pool state. we can split some of the work here by testing the pool state here. then, fuse that with the old-epoch spendable to apply cross-epoch constraints and perform the epoch lift.

making this a single step would means losing any proof assurances about at least one of those pcd inputs. the pool state might be a decent candidate in https://github.com/tachyon-zcash/tachyon/pull/59 but i avoided trying to justify that here.

i don't think there's any potential race conditions for *this* multi-step `SpendableEpochLift`. by the time this is happening in the proof system, all the events have landed on-chain, and we're just explaining history to the pcd.

if we have to split up these steps: if we can maintain the same conceptual structure, and conceal any decomposition behind existing headers, this logic should hold.

since `SpendableLift` and `StampLift` are simpler single-step fuses, i think those are also less likely to have any issues. again, with work on #59 we might be able to stop relying on pcd for the pool state input.

TalDerei: thanks for the in depth response, will spend some more time thinking about this.

turbocrime: to clarify concerns about race conditions and atomicity:

the thing we're mutating (pcd header to same pcd header) has to make an atomic transition in a single step.

the information we need to achieve that mutation might be safe to incrementally collect. we should try and be sure that intermediate headers:
- are purpose-constrained (unique header)
- prove something that is constrained by existing commitments
- prove something that is publicly testable later

book/src/proof-pipeline.md
turbocrime: the nullifier overlap check is correct if we say tx authors should lift their spendables to some height within the present epoch when they produce a stamp

PR #47, opened 2026-03-27
batch verifier

S1nus:
Batch verifier, similar API as orchard

- builds on https://github.com/tachyon-zcash/tachyon/pull/49 and https://github.com/tachyon-zcash/tachyon/pull/50

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -162,6 +162,7 @@ dependencies = [
  "pasta_curves",
  "rand",
  "rand_core",
+ "serde",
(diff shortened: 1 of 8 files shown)


TalDerei:
ragu equivalent, https://github.com/tachyon-zcash/ragu/issues/194.

turbocrime:
i think until there's something that tachyon can 'forward' from ragu, it makes more sense to mock this in consumers than to pretend we've got the feature

PR #54, opened 2026-04-09
rm serde, add zcash serialization

turbocrime:
Base: #45 

Removes `serde` dependency and `#[derive(Serialize, Deserialize)]` from all types. Adds consensus wire-format serialization for bundles using `zcash_encoding::CompactSize` and `core2::io::{Read, Write}`.

### Serialization

- `Stamped::read`/`write` and `Stripped::read`/`write` implement the consensus wire format
- `stampTachyon` field doubles as discriminant: values <= 0xFC indicate a stripped bundle with that stamp index; values >= 253 indicate a stamped bundle with a proof trailer
- `serialization.rs` module with shared helpers: `read_fp`/`write_fp`, `read_fq`/`write_fq`, `read_ep_affine`/`write_ep_affine`, `read_action_vk`/`write_action_vk`, `read_action_sig`/`write_action_sig`
- `Anchor::read`/`write`, `Tachygram::read`/`write`

### Bundle plan API

- `bundle::Plan::sign()` derives per-action signing keys and produces `Bundle<Unproven>`
- `bundle::Plan::apply_signatures()` accepts externally-produced signatures (e.g. FROST)
- `SignError` enum for sign/apply failures

### Stamp state model

- `Stampless` replaced by `Unproven` (signed but no proof) and `Adjunct` (stripped, with miner-assigned index)
- `Adjunct` carries a stamp index set by the miner during block assembly
- `Bundle<Unproven>::stamp()` attaches a proof, producing `Stamped`

### Removed

- `serde` dependency and feature flag
- `serde_helpers.rs`
- `witness.rs` / `ActionPrivate`
- `Plan::witness()` methods on `Plan<Spend>` and `Plan<Output>`
- `#[cfg_attr(feature = "serde", derive(...))]` from all types

### Dependencies

- Added: `core2`, `zcash_encoding`, `zcash_primitives`
- Removed: `serde`


crates/tachyon/src/bundle.rs
```diff
@@ -235,15 +517,160 @@ impl Stamped {
...
+    /// ### Stamp trailer
+    ///
+    /// If `stampTachyon` is not a single-byte value, a stamp trailer follows.
+    ///
+    /// | Name               | Format              | Description                       |
+    /// | ------------------ | ------------------- | --------------------------------- |
+    /// | `anchorTachyon`    | 32                  | pool anchor for this proof        |
+    /// | `nTachygrams`      | compactsize         | number of tachygrams              |
+    /// | `vTachygrams`      | 32 * nTachygrams    | tachygrams for this proof         |
+    /// | `proofTachyon`     | stampTachyon bytes  | a serialized proof, ~23 kilobytes |
```
turbocrime, line 556: block format including positional aggregate reference

crates/tachyon/src/bundle.rs
```diff
@@ -235,15 +517,160 @@ impl Stamped {
...
+    /// When finalizing a block, a miner will select some transaction order and
+    /// thus determine the position of each transaction containing a stamp.
+    /// Before serialization, stripped adjuncts should be provided an index
+    /// referring to their associated aggregate by its position among the stamps
+    /// in block order (only counting stamped bundles; this is not a 0-indexed
+    /// position among all transactions).
+    ///
+    /// Use of a single-byte compactsize value for this purpose technically
+    /// requires aggregate stamps to be located among the first 253 stamps in
+    /// the block. Given the present block limit of 2MB, this is acceptable.
```
turbocrime, line 652: some more detail on positional aggregate reference

crates/tachyon/src/bundle.rs
turbocrime: we could possibly use some high values of a single-byte `stampTachyon` as sentinel values, such as for 'no stamp'

for instance, we should probably never have a proof size greater than u16, so a 0xFE or 0xFF prefix could be a sentinel. or, to reduce parsing complexity, maybe we would just want to use high values like 0xFA and 0xFB that are unlikely to ever be real indexes.

turbocrime: an 'empty bundle' might have

- `nActionsTachyon == 0`
- `valueBalanceTachyon == 0`
- `bindingSigTachyon` signature over the sighash with a 0 bsk
- `stampTachyon == 0`, which looks like a stripped transaction referring to the 0th bundle as an aggregate, but an empty bundle would contribute no action digests to the aggregate

turbocrime: if we condition the presence of the rest of the bundle on `nActionsTachyon != 0` this precludes 'innocent' aggregates

if we add a flags field this makes the rest of parsing more complex

i'm inclined to just require non-tachyon transactions to sign an empty bundle with `stampTachyon == 0`

S1nus (approved):

PR #56, opened 2026-04-14
GGM refactor and `NoteId`

turbocrime:
## Summary

Rewrite the GGM tree PRF module around concrete key types and lazy recursive delegation, and introduce `NoteId` identity binding. No changes to the cryptographic tree structure or domain separators.

This is preliminary work for #46 

### Concrete key types

Replace the generic `NoteKey<D>` (parameterized over sealed `Master`/`Prefixed` depth markers) with concrete `NoteMasterKey` and `NotePrefixedKey` structs. This removes the `GGMTreeDepth` sealed trait, `PrefixedError` enum, and all depth-type machinery from `keys/note.rs`. Serialization moves onto the concrete types in `ggm.rs`.

### Lazy recursive delegation

Replace eager dyadic decomposition (`Prefixed::tight(start, end)`) with `derive_note_delegates(RangeInclusive<u32>)` on both key types. Only tree nodes overlapping the requested range are hashed — no up-front interval arithmetic.

### Fluent tree traversal

Add `step(bool)` on both `NoteMasterKey` and `NotePrefixedKey` for single-level descent, and `range()` on `NotePrefixedKey` to recover the covered epoch interval.

### NoteId

New `NoteId` primitive: `H(domain, mk, cm)` binds a note's identity (via its master key) to its commitment. `NullifierKey::note_id(&Note)` derives both `mk` and `cm` internally so callers can't misassociate them. Domain tag: `Tachyon-NoteMkCm`.

### Behavioural changes

- `NotePrefixedKey::derive_nullifier` panics on out-of-range epochs (previously returned `Option`). Callers already know their range.
- `GGM_TREE_DEPTH` is now `pub` (was `pub(super)`).

## Files changed

| File | Change |
|---|---|
| `keys/ggm.rs` | Full rewrite: concrete types, recursive delegation, `step`/`range` API, serialization |
| `keys/note.rs` | Remove `NoteKey<D>` and depth-type machinery; add `note_id()` accessor |
| `keys/mod.rs` | Re-export `GGM_TREE_DEPTH`, `NoteMasterKey`, `NotePrefixedKey` from `ggm` |
| `primitives/note_id.rs` | New `NoteId` type with `derive(nk, note)` |
| `primitives/mod.rs` | Wire `NoteId` |
| `primitives/effect.rs` | Docstring rewrap (no logic change) |
| `constants.rs` | Add `NOTE_ID_DOMAIN` |

### DIFF
--- a/crates/tachyon/src/keys/ggm.rs
+++ b/crates/tachyon/src/keys/ggm.rs
@@ -57,23 +57,17 @@ impl NoteMasterKey {
     /// only hashing children that overlap the range.
     #[must_use]
     pub fn derive_note_delegates(&self, range: RangeInclusive<u32>) -> Vec<NotePrefixedKey> {
-        let (low, high) = {
-            #[expect(clippy::expect_used, reason = "guaranteed valid shift")]
-            let root_split = u32::MAX
-                .checked_shr(1u32)
-                .expect("shifting u32::MAX by 1 will never overflow");
-            (
-                *range.start()..=root_split.min(*range.end()),
-                (*range.start()).max(root_split + 1)..=*range.end(),
-            )
-        };
+        // Children at depth 1 each own half of the u32 epoch space.
+        let split: u32 = u32::MAX >> 1;
 
         let mut result = Vec::new();
-        if !low.is_empty() {
-            result.extend(self.step(false).derive_note_delegates(low));
+        if *range.start() <= split {
+            let lo = *range.start()..=(*range.end()).min(split);
+            result.extend(self.step(false).derive_note_delegates(lo));
         }
-        if !high.is_empty() {
-            result.extend(self.step(true).derive_note_delegates(high));
+        if *range.end() > split {
+            let hi = (*range.start()).max(split + 1)..=*range.end();
+            result.extend(self.step(true).derive_note_delegates(hi));
         }
         result
     }
@@ -150,32 +144,36 @@ impl NotePrefixedKey {
     ///
     /// Recursively descends the tree, emitting fully-covered nodes and
     /// only hashing children that overlap the range.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `range` is not a subset of [`Self::range`].
     #[must_use]
     pub fn derive_note_delegates(&self, range: RangeInclusive<u32>) -> Vec<Self> {
+        assert!(
+            self.range().contains(range.start()) && self.range().contains(range.end()),
+            "prefix key for {:?} does not cover requested range {:?}",
+            self.range(),
+            range,
+        );
+
         if range == self.range() {
             // This node exactly covers the requested range.
             alloc::vec![*self]
         } else {
             // This node is larger than the requested range.
-            let (low, high) = {
-                let next_depth = u32::from(self.depth.get() + 1);
-                let next_dyad_size = u32::MAX.checked_shr(next_depth).unwrap_or(0);
-
-                // a key's range is dyadic-aligned. select next smallest dyad.
-                let next_split = *self.range().start() | next_dyad_size;
-
-                (
-                    *range.start()..=next_split.min(*range.end()),
-                    (*range.start()).max(next_split + 1)..=*range.end(),
-                )
-            };
+            let next_depth = u32::from(self.depth.get() + 1);
+            let next_dyad_size = u32::MAX.checked_shr(next_depth).unwrap_or(0);
+            let split = *self.range().start() | next_dyad_size;
 
             let mut result = Vec::new();
-            if !low.is_empty() {
-                result.extend(self.step(false).derive_note_delegates(low));
+            if *range.start() <= split {
+                let lo = *range.start()..=(*range.end()).min(split);
+                result.extend(self.step(false).derive_note_delegates(lo));
             }
-            if !high.is_empty() {
-                result.extend(self.step(true).derive_note_delegates(high));
+            if *range.end() > split {
+                let hi = (*range.start()).max(split + 1)..=*range.end();
+                result.extend(self.step(true).derive_note_delegates(hi));
             }
             result
         }
@@ -341,5 +339,41 @@ mod tests {
         }
         // Now at depth 32 (leaf) — one more step should panic.
         let _boom = key.step(false);
+    }
+
+    #[test]
+    fn full_range_from_master() {
+        let root = NoteMasterKey(Fp::from(1u64));
+        let delegates = root.derive_note_delegates(0..=u32::MAX);
+        assert_eq!(delegates.len(), 2);
+        assert_eq!(delegates[0].range(), 0..=(u32::MAX >> 1u32));
+        assert_eq!(delegates[1].range(), (u32::MAX >> 1u32) + 1..=u32::MAX);
+    }
+
+    #[test]
+    fn last_epoch_delegate() {
+        let root = NoteMasterKey(Fp::from(1u64));
+        let delegates = root.derive_note_delegates(u32::MAX..=u32::MAX);
+        assert_eq!(delegates.len(), 1);
+        assert_eq!(delegates[0].range(), u32::MAX..=u32::MAX);
+        assert_eq!(delegates[0].depth.get(), GGM_TREE_DEPTH);
+    }
+
+    #[test]
+    #[should_panic(expected = "does not cover requested range")]
+    fn disjoint_range_panics() {
+        let root = NoteMasterKey(Fp::from(1u64));
+        // depth-2 prefix covering [0..=0x3FFF_FFFF].
+        let prefix = root.step(false).step(false);
+        let _delegates = prefix.derive_note_delegates(0x8000_0000..=0x8000_0010);
+    }
+
+    #[test]
+    #[should_panic(expected = "does not cover requested range")]
+    fn partial_overlap_panics() {
+        let root = NoteMasterKey(Fp::from(1u64));
+        // depth-2 prefix covering [0..=0x3FFF_FFFF].
+        let prefix = root.step(false).step(false);
+        let _delegates = prefix.derive_note_delegates(0..=0x4000_0000);
     }
 }
--- a/crates/tachyon/src/keys/ggm.rs
+++ b/crates/tachyon/src/keys/ggm.rs
@@ -6,197 +6,268 @@
 //! leaves, enabling contiguous-range prefix delegation.
 
 use alloc::vec::Vec;
-use core::num::NonZeroU8;
+use core::{num::NonZeroU8, ops::RangeInclusive};
 
 use ff::PrimeField as _;
 // TODO(#39): replace halo2_poseidon with Ragu Poseidon params
 use halo2_poseidon::{ConstantLength, Hash, P128Pow5T3};
 use pasta_curves::Fp;
 
-use super::note::NoteKey;
 use crate::{constants::NOTE_NULLIFIER_DOMAIN, note::Nullifier, primitives::Epoch};
 
-/// GGM tree depth — 32-bit epochs cover ~4 billion values.
-pub(super) const MAX_TREE_DEPTH: u8 = 32;
+/// GGM tree depth — 32-bit epochs, leaves at depth 32.
+pub const GGM_TREE_DEPTH: u8 = 32;
 
-/// Marker for a master (root, depth 0) note key.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub struct Master;
-
-/// A GGM subtree identified by its depth and node index.
+/// Per-note master root key.
 ///
-/// At depth `d` there are `2^d` nodes. Node `i` covers the contiguous epoch
-/// range `[i * 2^(32-d) ..= (i+1) * 2^(32-d) - 1]`. The index encodes the
-/// `d`-bit path from root (MSB-first).
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
-pub struct Prefixed {
-    depth: NonZeroU8,
-    index: u32,
-}
+/// Root of the GGM tree PRF for a single note. Derived by the user device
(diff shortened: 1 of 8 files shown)


crates/tachyon/src/keys/ggm.rs
```diff
@@ -6,317 +6,340 @@
...
 use halo2_poseidon::{ConstantLength, Hash, P128Pow5T3};
 use pasta_curves::Fp;
 
-use super::note::NoteKey;
 use crate::{constants::NOTE_NULLIFIER_DOMAIN, note::Nullifier, primitives::Epoch};
 
-/// GGM tree depth — 32-bit epochs cover ~4 billion values.
-pub(super) const MAX_TREE_DEPTH: u8 = 32;
+/// GGM tree depth — 32-bit epochs, leaves at depth 32.
+pub const GGM_TREE_DEPTH: u8 = 32;
```
TalDerei, line 19: ACK – references https://github.com/tachyon-zcash/tachyon/pull/28#discussion_r2920719274. this is likely overkill, but we can revisit the constant.

crates/tachyon/src/keys/note.rs
```diff
@@ -121,17 +44,27 @@ impl NullifierKey {
     ///   F_{\mathsf{mk}}(\text{flavor})$
     /// - Derive epoch-restricted prefix keys $\Psi_t$ for OSS delegation
     #[must_use]
-    pub fn derive_note_private(&self, psi: &NullifierTrapdoor) -> NoteMasterKey {
+    pub fn derive_note_private(&self, psi: &note::NullifierTrapdoor) -> NoteMasterKey {
```
TalDerei, line 47: ACK – poseidon KDF that creates `NoteMasterKey` per note.

crates/tachyon/src/keys/ggm.rs
```diff
@@ -6,317 +6,340 @@
...
-    /// Last leaf index in the covered range (inclusive).
-    pub const fn last(self) -> u32 {
-        self.first() | (u32::MAX >> self.depth.get())
-    }
+    /// Derive epoch-restricted prefix keys covering the specified range.
+    ///
+    /// Recursively descends the tree, emitting fully-covered nodes and
+    /// only hashing children that overlap the range.
+    #[must_use]
+    pub fn derive_note_delegates(&self, range: RangeInclusive<u32>) -> Vec<NotePrefixedKey> {
```
TalDerei, line 59: ACK – `NoteMasterKey::derive_note_delegates` lazily emits minimal dyadic covers as `NotePrefixedKey` for OSS handoff.

crates/tachyon/src/keys/ggm.rs
```diff
@@ -6,317 +6,340 @@
...
-            .map(|prefix| {
-                NoteKey {
-                    inner: walk(self.inner, prefix.index, prefix.depth()),
-                    prefix,
-                }
-            })
-            .collect()
+    /// Panics if the epoch is outside this key's authorized range.
+    #[must_use]
+    pub fn derive_nullifier(&self, flavor: Epoch) -> Nullifier {
```
TalDerei, line 188: ACK – OSS uses the delegation key to walk `GGM_TREE_DEPTH - dk.depth`. I'll blackbox the actual `ggm_walk` logic with the shared understanding that this is sound if the delegation key _can't_ hash outside it's root (only within its subtree, traversing downward — saw the `Option` change, but maybe it should be fallible with a `Result`?).

haven't thought about what i'll loosely call the "OSS wiring protocol" for how it determines which GGM tree to traverse for which note (assuming `NoteId` is the OSS handle there)? like the OSS needs some kind of crude indexing like `table[note_id] = delegates` in a DB.

crates/tachyon/src/keys/ggm.rs
```diff
@@ -6,317 +6,340 @@
...
-        self.index
     }
 
-    /// First leaf index in the covered range.
-    pub const fn first(self) -> u32 {
-        let height = MAX_TREE_DEPTH - self.depth.get();
-        self.index << height
+    /// Derive a nullifier for the given epoch.
+    #[must_use]
+    pub fn derive_nullifier(&self, flavor: Epoch) -> Nullifier {
```
TalDerei, line 50: ACK – fair, user should be able to derive nullifiers and run their own OSS service in theory.

crates/tachyon/src/primitives/note_id.rs
```diff
@@ -0,0 +1,43 @@
...
+
+use crate::{constants::NOTE_ID_DOMAIN, keys, note};
+
+/// Identity binding: `H(domain, mk, cm)`.
+///
+/// Binds the note's identity (via master key) to its value commitment,
+/// threading through the proof pipeline. Fuse steps verify that left
+/// and right inputs agree on note id.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct NoteId(Fp);
```
TalDerei, line 13: this is _probably_ the right shape for our context, would need to think about it more.

crates/tachyon/src/keys/note.rs
```diff
@@ -163,34 +96,15 @@ impl NullifierKey {
...
-/// ## Delegation chain
-///
-/// ```text
-/// nk + psi → mk (per-note root, user device)
-///              ├── nf = F_mk(flavor)     nullifier for a specific epoch
-///              └── psi_t = GGM(mk, t)    prefix key for epochs e ≤ t (OSS)
-/// ```
-///
-/// `mk` is not stored or transmitted — the user device derives it
-/// ephemerally when needed. The OSS receives only the prefix keys.
```
TalDerei, line 180: nit; keep language on mk lifetime.

crates/tachyon/src/keys/ggm.rs
```diff
@@ -6,317 +6,340 @@
...
-    /// F_{\mathsf{mk}}(\text{flavor})$.
+    /// Derive epoch-restricted prefix keys covering the specified range
+    /// within this key's range.
+    ///
+    /// Recursively descends the tree, emitting fully-covered nodes and
+    /// only hashing children that overlap the range.
     #[must_use]
-    pub fn derive_nullifier(&self, flavor: Epoch) -> Nullifier {
-        Nullifier::from(self.evaluate(u32::from(flavor)))
+    pub fn derive_note_delegates(&self, range: RangeInclusive<u32>) -> Vec<Self> {
```
TalDerei, line 152: why does `NotePrefixedKey::derive_note_delegates` exist? I don't see a reason to inductively derive tighter sub-delegates and this is probably buggy.

crates/tachyon/src/keys/ggm.rs
```diff
@@ -6,317 +6,340 @@
...
+///
+/// ## Delegation chain
+///
+/// ```text
+/// nk + psi → mk (per-note root, user device)
+///              ├── nf = F_mk(flavor)     nullifier for a specific epoch
+///              └── psi_t = GGM(mk, t)    prefix key for epochs e ≤ t (OSS)
+/// ```
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct NoteMasterKey(pub Fp);
```
TalDerei: I'd like to tighten up visibility per https://github.com/tachyon-zcash/tachyon/pull/56/changes#r3096760532, limiting external exposure. 

```ts
pub struct NoteMasterKey(pub(crate) Fp);
```

crates/tachyon/src/keys/ggm.rs
```diff
@@ -6,317 +6,340 @@
...
+pub struct NotePrefixedKey {
+    /// GGM tree node value.
+    pub inner: Fp,
     /// The number of levels already descended.
-    pub fn depth(self) -> u8 {
-        self.prefix.depth()
+    pub depth: NonZeroU8,
+    /// Node index at this depth.
+    pub index: u32,
+}
```
TalDerei, line 107: let's make the internal fields `pub(crate)` to force callers to go through `TryFrom`.

crates/tachyon/src/keys/ggm.rs
```diff
@@ -6,317 +6,340 @@
...
+/// from [`NullifierKey`](super::NullifierKey) and the note's psi trapdoor.
+///
+/// ## Delegation chain
+///
+/// ```text
+/// nk + psi → mk (per-note root, user device)
+///              ├── nf = F_mk(flavor)     nullifier for a specific epoch
+///              └── psi_t = GGM(mk, t)    prefix key for epochs e ≤ t (OSS)
+/// ```
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
```
TalDerei, line 33: we should add custom debugs to avoiding leaking secrets (think fmt prints the full canonical field element).

crates/tachyon/src/primitives/note_id.rs
```diff
@@ -0,0 +1,43 @@
...
+            ]),
+        )
+    }
+}
+
+impl From<Fp> for NoteId {
+    fn from(fp: Fp) -> Self {
+        Self(fp)
+    }
+}
```
TalDerei, line 38:
```suggestion

impl From<NoteId> for Fp {
    fn from(id: NoteId) -> Self {
        id.0
    }
}

File: crates/tachyon/src/keys/note.rs
Comment (TalDerei) [inline crates/tachyon/src/keys/note.rs:1]: let's make `pub psi: NullifierTrapdoor` and `pub rcm: CommitmentTrapdoor` as `pub(crate)` in this file. we should do a holistic visibility pass of sensitive internal fields elements.

File: crates/tachyon/src/primitives/note_id.rs [hunk 13]
```diff
@@ -0,0 +1,43 @@
...
+/// Identity binding: `H(domain, mk, cm)`.
+///
+/// Binds the note's identity (via master key) to its value commitment,
+/// threading through the proof pipeline. Fuse steps verify that left
+/// and right inputs agree on note id.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct NoteId(Fp);
+
+impl NoteId {
+    pub(crate) fn derive(nk: &keys::NullifierKey, note: &note::Note) -> Self {
```
turbocrime, line 17: it's been suggested that NoteId derivation might involve a trapdoor so it's not so deterministic

crates/tachyon/src/keys/ggm.rs
turbocrime: it's recursive, and this is the normal path.

in practice, all derivation is going to happen inside a ragu step, and all of this will be obviated.

implementing it recursively like this is more realistic than any other approach we could take right now.

crates/tachyon/src/keys/ggm.rs
turbocrime: fields now pub(crate)

crates/tachyon/src/keys/note.rs
turbocrime: i think in general changing visibility is good for encouraging proper handling, but can also be misleading in that it means the presence of sensitive material may be non-obvious

making key fields pub(crate): fine. it's always obvious when you're handling a key, it has key in the name.

making a selection of note fields private: disagree. it is no longer clear you're handling an object containing real secrets.

crates/tachyon/src/keys/ggm.rs
turbocrime: it feels a little 'early' to be doing that but ok. pushed a commit changing field visibility.

crates/tachyon/src/keys/ggm.rs
turbocrime: i think this deserves to be its own issue as deciding what falls in/out of this scope of concern is complex

turbocrime: https://github.com/tachyon-zcash/tachyon/issues/60

crates/tachyon/src/primitives/note_id.rs
turbocrime: it was suggested in tuesday sync that we might want to add a trapdoor so that the note id can be randomized. that relationship becomes more private, and proofs can open it by witnessing the trapdoor

crates/tachyon/src/keys/ggm.rs
turbocrime: anyone knowing the root key material can check your derivation. it's not possible to derive correctly outside of your given scope.

but of course most of consensus does not know your root key material. these are just functions, and you can always just call them on some inputs to get some numbers.

the step implementation in #46 constrains derivation and scope and key relationships.

turbocrime: yes note_id is the oss handle. anyone using a prefixed key will be identifying things by note_id, NOT note commitment, and testing whatever nullifiers pop out the other end of the tree.

crates/tachyon/src/keys/note.rs
turbocrime: it's deleted because i'm not confident it's correct, or good advice.

crates/tachyon/src/keys/ggm.rs
TalDerei: unresolving; intuition suggested running this method's logic through claude, and it found some kind of edge case bug. can you investigate some kind of simplification?

crates/tachyon/src/primitives/note_id.rs
TalDerei: yeah remembering that thread now, can we add a comment to that effect?

TalDerei:
https://github.com/tachyon-zcash/tachyon/pull/56#discussion_r3100250007

crates/tachyon/src/keys/ggm.rs
turbocrime: improved https://github.com/tachyon-zcash/tachyon/pull/56/commits/7d9d463fa81c48163284520cf16689d677de1995

panic on incorrect use is deliberate. now, it fails early with an assertion. this covers the edge cases.

TalDerei (approved):

PR #57, opened 2026-04-16
Correct payment key derivation

turbocrime:
Base https://github.com/tachyon-zcash/tachyon/pull/56

## Summary

- Derive `pk` from `(ak, nk)` via Poseidon instead of directly from `sk` via PRF, matching original intent
- Add `PaymentKey::derive(ak, nk)` and `ProofAuthorizingKey::derive_payment_key()`
- Remove `PrfExpand::PK`

## Motivation

The previous derivation (`pk = ToBase(PRF(sk, 0x0b))`) made `pk` independent of both `ak` and `nk`. This breaks the cryptographic binding between the payment key committed in notes and the proof authorizing key used for nullifier derivation.

Then, because `cm = Poseidon(pk, v, psi, rcm)` didn't relate to an `nk`, an arbitrary `nk` might 'prove' derivation of unrelated nullifiers for any given `cm`.

## Solution

Derive `pk = Poseidon(PK_DOMAIN, ak_x, nk)`. Since `pk` is already committed in `cm`, note commitments now transitively bind to both `ak` and `nk`, and proof steps may properly constrain note data.

### DIFF
--- a/crates/tachyon/src/keys/mod.rs
+++ b/crates/tachyon/src/keys/mod.rs
@@ -90,7 +90,7 @@ mod tests {
     use crate::{
         constants::PrfExpand,
         entropy::ActionEntropy,
-        keys::private,
+        keys::{NullifierKey, PaymentKey, private},
         note::{self, CommitmentTrapdoor, Note, NullifierTrapdoor},
         primitives::effect,
         reddsa,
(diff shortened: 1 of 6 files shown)


crates/tachyon/src/keys/note.rs
```diff
@@ -96,6 +108,26 @@ impl NullifierKey {
...
 
+impl PaymentKey {
+    /// Derive the payment key from `ak` and `nk`:
+    /// $\mathsf{pk} = \text{Poseidon}(\text{PK\_DOMAIN}, \mathsf{ak}_x, \mathsf{nk})$.
+    #[must_use]
+    #[expect(
+        clippy::expect_used,
+        reason = "sign-normalized ak (tilde_y=0) is always a valid Fp repr"
+    )]
+    pub fn derive(ak: &SpendValidatingKey, nk: &NullifierKey) -> Self {
```
TalDerei, line 121: ACK — was expecting this key derivation change. `cm` transitively binds to `nk` via `pk`, but we'd need corresponding checks in the dummy spend path in `stamp/proof.rs` to enforce this fully obviously. we can just add a note there.

crates/tachyon/src/keys/mod.rs
```diff
@@ -129,17 +131,21 @@ mod tests {
         assert_eq!(flipped, 16u32);
     }
 
-    /// ask, nk, pk derived from the same sk must all be different
-    /// (different domain separators produce independent keys).
+    /// ask, nk, pk derived from the same sk must all be different.
+    /// pk derives from (ak, nk) via Poseidon, not directly from sk.
     #[test]
     fn child_keys_independent() {
```
TalDerei, line 137: test doesn't exercise the binding where varying `nk` changes `pk`.

TalDerei (approved):
strictly superior, we can iterate.

crates/tachyon/src/keys/note.rs
turbocrime: resolving as handled by https://github.com/tachyon-zcash/tachyon/pull/46

crates/tachyon/src/keys/mod.rs
turbocrime: adding test

PR #58, opened 2026-04-16
mock_ragu: add `Multiset` gadget

turbocrime:
Adds a `Multiset` gadget mocking a future ragu API for a queryable committed polynomial: `commit` / `commit_with(blind)`, `query(point)` (direct evaluation standing in for a PCS opening claim), and `merge` (polynomial multiplication, double-counting shared elements). Equality compares commitments.

Also, tests reorganized from a single flat file into per-module test files.


crates/mock_ragu/src/tests/application.rs
turbocrime, line 1: i split the monolithic tests.rs into multiple files. github identified one of them as 'relocated', so all of these removals you see are just moved to other test files.

PR #61, opened 2026-04-20
accumulator

S1nus:
Stores the height and state of the accumulator as a `u64` and `Fq` element, obtained from the Poseidon hash of the concatenation of the previous accumulator state and a pedersen commitment to the coefficients of a polynomial with roots at the accumulated Fp values.


TalDerei:
discussed OOB to close in favor of impl in https://github.com/tachyon-zcash/tachyon/pull/46?

PR #62, opened 2026-04-27
Late delegation blinding for GGM derivation

turbocrime:
splits the GGM tree descent into two paths

- Private descent from the note master key, carrying `(mk, cm)`
- Blinded descent from some prefix key, carrying `delegation_id`


book/src/nullifier-derivation.md
```diff
@@ -24,18 +24,18 @@ Dan Boneh and Brent Waters introduced the "Constrained Pseudorandom Functions an
...
+Let $F_K$ be a $k$-ary GGM tree PRF instantiated from Poseidon (P128Pow5T3, domain tag `Tachyon-NfDerive`). Each tree step computes $\text{Poseidon}(\text{tag}, \text{node}, \text{chunk})$, where *chunk* is a `LOG2_ARITY`-bit slice of the epoch index. Tachyon uses `LOG2_ARITY = 2` (so $k = 4$), and sizes the tree to tile exactly the epoch space: `GGM_TREE_DEPTH = EPOCH_BITS / LOG2_ARITY`, where `EPOCH_BITS = u32::BITS - EPOCH_SHIFT`. The wallet derives the master root key as $mk = \text{Poseidon}_\text{Tachyon-MkDerive}(\psi, nk)$.
 
-1. **Compute the minimal prefix cover of [0..t].**
+1. **Compute a prefix cover of [s..=t].**
 
-   Decompose the integer range into power-of-two aligned subranges (dyadic intervals), and each subrange corresponds to a binary prefix over epoch bits.
+   Start from the epoch $s$ of the spendable the wallet already holds (see "Delegation window" below); $s$ ranges from $0$ (first contact with sync) to the last-advanced spendable epoch. Decompose $[s..=t]$ into $k$-ary aligned subranges (each subrange is $k^j$ consecutive epochs on a $k^j$-aligned boundary for some $j$). Each subrange corresponds to a base-$k$ prefix over epoch chunks.
```
TalDerei: this is an interesting design decision; I had thought about k-arity mental model early on, but didn't have a good intuition for the cost trade-off between delegation key bandwidth and in-circuit hash costs (what the scaling factor there was). so is it an empirical finding that a quaternary-width tree optimizes those costs?

book/src/nullifier-derivation.md
```diff
@@ -61,6 +61,104 @@ The Oblivious Syncing Service (**OSS**) is granted key capability that allows it
...
+prefix, and the prefix's nullifier leaf) and reuse them across delegations.
+A delegation event is then a single in-circuit step that attaches a fresh
+trapdoor to a cached `NfPrefixHeader`.
+
+The chain has two leaves: a **pre-blind** `NullifierHeader` produced on the
+user device, carrying $(cm, nf, epoch)$ for `SpendBind` cm-equality and
+`SpendableInit` pool-membership; and a **post-blind**
+`DelegateNullifierHeader` produced by the sync service, carrying
+$(nf, epoch, \text{delegation\_id})$ for `SpendableRollover` cross-epoch
+matching.
```
TalDerei: another interesting design decision; this caching is obvious in hindsight because on shared paths the pre-blind prefix PCDs can reused, but this also implies to me that for non-overlapping old paths there'd need to be some wallet garbage collection mechanism there, which is fine. this symmetry between the wallet and OSS both converging on the same leaf (same nullifier) but using different PCD chains is really appealing.

book/src/nullifier-derivation.md
```diff
@@ -61,6 +61,104 @@ The Oblivious Syncing Service (**OSS**) is granted key capability that allows it
...
+can match nullifiers at overlapping epochs to confirm that both
+delegations concern the same note. A tight $[s..=t]$ cover with $s$
+equal to the held spendable's epoch minimises this overlap, since prior
+delegations covered epochs $\leq s$ and the new one covers $\geq s$.
+Widening with $s' < s$ re-exposes epochs covered by earlier delegations.
+The degenerate widening $s' = 0$ reproduces the original $[0..=t]$
+behavior — maximum linkability risk — as the first list entry.
+
+Spine caching shaves the leftmost-peak chain, so covers that concentrate
+peaks on the spine benefit most in effort terms.
```
TalDerei: okay so enumerating the privacy leaks here:
- same delegation event to two OSS (same `trapdoor / delegation_id / prefix keys` and observing any one of those mean trivial likability)
- different delegation events to two OSS (fresh trapdoor, so `delegation_id` unlinkable, but prefix keys and nf are linkable if the two covers share an aligned subtree or two covers share epochs, respectively).

this seems fine for now until that reblind step you hinted at provides presumingly full unlinkability?

book/src/nullifier-derivation.md
```diff
@@ -61,6 +61,104 @@ The Oblivious Syncing Service (**OSS**) is granted key capability that allows it
...
+caller measures derivation effort on each candidate via
+`derive_note_delegates(range)` and picks by its own policy weighing
+effort against linkability exposure.
+
 ## Assumptions
 
-This currently assumes a fixed-depth tree.
+Current parameters: `EPOCH_SHIFT = 12` (production) / `4` (test), `LOG2_ARITY = 2`.
+`GGM_TREE_DEPTH` is derived, not configured independently, so the tree
+structure always matches the reachable epoch space.
```
TalDerei: with the [proposal for lowering block times](https://forum.zcashcommunity.com/t/proposal-lower-zcash-block-target-spacing-to-25s/54577), whatever this blocks-per-epoch ratio is should probably be increased to favor even longer epochs times, which is my intuition.

book/src/nullifier-derivation.md
TalDerei, line 1: general book observation not associated with this entry; our nullifier derivation and OSS impl imply tight covers + periodically-online wallet assumptions, but we should also document the other case where the wallet basically is offline for a long time, and later comes back and wants to spend their funds. 

obviously the OSS can only perform continuous spendabilty maintenance (while the wallet is offline) only the current prefix cover for epoch T that wallet delegated. say the wallet goes offline for a long time and X epochs pass and chain tip advances to T', then there will need be some redelegation event to update the cover. we're also assuming here that the user can't delegate epochs unless they're closed on-chain, since delegating future epochs is a privacy leak and we probably shouldn't hand the user such a footgun, and there should probably be some kind of guard for preventing future-epoch delegations. maybe the guard is just the protocol also choosing the latest epoch and there isn't like a wallet configurability knob.

anyway, the case here is when OSS is idle for a particular note because the wallet was offline for some an extended period and the chain tip extends beyond the prefix cover, and then the wallet coming online will incur some delay before spending their funds, which is expected but should be documented. 

also here's another idea; maybe at the wallet layer, there should be some background task that periodically kicks off a redelegation event (recalling https://github.com/prax-wallet/prax/pull/269).

crates/tachyon/src/stamp/proof/delegation.rs
```diff
@@ -1,96 +1,173 @@
...
 };
 
-/// Header for the note master key.
+/// Pre-blind header for the note master key.
+///
+/// Carries the `(mk, cm)` lineage established at [`NfMasterSeed`]. No
+/// delegation identifier — blinding happens later.
 #[derive(Clone, Debug)]
-pub struct DelegationMasterHeader;
+pub struct NfMasterHeader;
```
TalDerei, line 46: nit; these pre-blind PCDs (`NfMasterHeader`, `NfPrefixHeader`, `NullifierHeader`) expose `mk` and `cm` as pub data on `Pcd<H>`, so it'd be nice if the type system enforced that they can never accidentally reach the OSS (like some kind of marker trait).

crates/tachyon/src/stamp/proof/delegation.rs
```diff
@@ -102,72 +179,178 @@ impl Step for DelegationSeed {
...
+        let domain = Fp::from_u128(u128::from_le_bytes(*DELEGATION_ID_DOMAIN));
+        let delegation_id = DelegationId::from(
+            &Hash::<_, P128Pow5T3, ConstantLength<4>, 3, 2>::init().hash([
+                domain,
+                mk.0,
+                Fp::from(&cm),
+                Fp::from(&trap),
+            ]),
+        );
+        Ok(((key, delegation_id), ()))
```
TalDerei, line 309: nit; `delegation_id` derivation is duplicated in a few places, maybe this should be refactored into a shared helper?

crates/tachyon/src/stamp/proof/delegation.rs
```diff
@@ -1,96 +1,173 @@
...
 #[derive(Clone, Debug)]
-pub struct DelegationMasterHeader;
+pub struct NfMasterHeader;
 
-impl Header for DelegationMasterHeader {
-    type Data<'source> = (NoteMasterKey, DelegationId);
+impl Header for NfMasterHeader {
+    type Data<'source> = (NoteMasterKey, note::Commitment);
 
     const SUFFIX: Suffix = Suffix::new(0);
```
TalDerei, line 51: trivial () header already takes this suffix key, so there's a collision. also, noticed that `mock_ragu` doesn't hash suffix / index, so that mock impl diverges with ragu in that respect.

crates/tachyon/src/stamp/proof/spend.rs
```diff
@@ -53,42 +57,40 @@ impl Step for SpendBind {
...
     ) -> mock_ragu::Result<(<Self::Output as Header>::Data<'source>, Self::Aux<'source>)> {
-        if left_delegation_id != right_delegation_id {
+        // Same wallet: both leaves share the cm propagated from the
+        // pre-blind NfMasterSeed root.
+        if left_cm_tg != right_cm_tg {
             return Err(mock_ragu::Error);
         }
         if right_epoch.0 != left_epoch.0 + 1 {
             return Err(mock_ragu::Error);
         }
```
TalDerei, line 78: nit; defense-in-depth about these kinds of bounds wrapping:

```ts
  let Some(expected_next) = old_epoch.0.checked_add(1) else {
      return Err(mock_ragu::Error);
  };
  if new_epoch.0 != expected_next {
      return Err(mock_ragu::Error);
  }
```

crates/tachyon/src/stamp/proof/spend.rs
```diff
@@ -53,42 +57,40 @@ impl Step for SpendBind {
...
-        if u64::from(note.value) == 0 {
+        // Bind the witnessed note to the leaves' cm.
+        let note_cm_tg = Tachygram::from(&Fp::from(&note.commitment()));
+        if note_cm_tg != left_cm_tg {
             return Err(mock_ragu::Error);
         }
-        if note.pk.0 != pak.derive_payment_key().0 {
+        if u64::from(note.value) == 0 {
```
TalDerei: nit; what about upper-bound checks?

crates/tachyon/src/stamp/tests.rs
```diff
@@ -52,45 +52,31 @@ fn merge_stamp_rejects_mismatched_anchors() {
...
-        note_b,
-        trap_b,
-        anchor,
-        pool_state,
-        nf_now_b.clone(),
-    );
+    let master_a = user.note_master(&mut rng, note_a);
+    let master_b = user.note_master(&mut rng, note_b);
+    let (nf_now_a, nf_next_a) =
+        preblind_nullifier_pair_from_master(&mut rng, master_a, target_epoch);
```
TalDerei, line 75: was trying to figure out what this `preblind_nullifier_pair_from_master` did, and just tracing the logic, there _may_ be a domain-range overflow bug, can you investigate that? think that method should return a `Result`

```ts
      if target_epoch.0 >= GGM_MAX_INDEX {
          return Err();
      }
```

TalDerei (approved):
non-blocking observations for follow-up work 👍

book/src/nullifier-derivation.md
turbocrime: the wallet-side garbage collection should be as simple as 'do i have a spendable this old'

book/src/nullifier-derivation.md
turbocrime: at this point i think reblinding is not useful. the user will have a cached prefix nearby, and they can just blind a new delegation.

a user trying to maximize privacy will assume all delegations are colluding. so, they will not tolerate overlap in delegation. they must either

- get a new note (spend and output to self), or
- wait for the range to expire, privately perform epoch rollover, then delegate a disjoint range

at this point and in later PRs, delegation_id is only useful for delegates to perform epoch rollovers.

book/src/nullifier-derivation.md
turbocrime: sure. these are all tunable.

book/src/nullifier-derivation.md
turbocrime: yes, been thinking about these various ways to use a sync service. i think probably what's going to happen is most users will simply not update their spendables, and then delegate to a long-term history service when they actually want to spend.

this is really a question for wallet developers and it will be answered by sync service pricing for deep history vs constant maintenance.

i attempted to design tools flexible enough for the whole solution space. the important stuff for that (inclusion/exclusion testing, etc) is in the following PRs.

TalDerei: can you clarify how you're internalizing this long-term history service? like some kind of batch pay per catch-up request specifically for the long-term offline case?

crates/tachyon/src/stamp/proof/delegation.rs
turbocrime: the proof tree is structured to scope it out. user-side proofs take the private versions, delegate-side proofs take the delegate versions.

beyond that, it's up to consumers, who simply shouldn't publish their secrets. we can't actually write a type that enforces this.

this is essentially the same convo as https://github.com/tachyon-zcash/tachyon/pull/34#discussion_r2922321192 and https://github.com/tachyon-zcash/tachyon/pull/29 which deserves some revisit. we should definitely attempt to name things in ways that make it obvious how carefully to handle them.

crates/tachyon/src/stamp/proof/delegation.rs
turbocrime: this is done by #74

crates/tachyon/src/stamp/proof/spend.rs
turbocrime: i expect the step guts will be completely rewritten when we transition away from the mock, so i think this is fine right now. i might add a lint to enforce.

crates/tachyon/src/stamp/proof/spend.rs
turbocrime: ok, will add a check against NOTE_VALUE_MAX

book/src/nullifier-derivation.md
TalDerei: that's fair, filed https://github.com/tachyon-zcash/tachyon/issues/76 for tracking.

book/src/nullifier-derivation.md
TalDerei: filed https://github.com/tachyon-zcash/tachyon/issues/77 for tracking.

crates/tachyon/src/stamp/tests.rs
turbocrime: this is a test utility. it's also completely removed by #64 so i am not worried about it

book/src/nullifier-derivation.md
turbocrime: sorry, this is a bad draft. some of it appears incorrect.

the parameters are not selected empirically. they are extremely arbitrary.

i introduced parameters to provide knobs that we can tune when we have something more solid to reason about. also, the tree can shrink for the test walks.

TalDerei: filed https://github.com/tachyon-zcash/tachyon/issues/78 for tracking.

book/src/nullifier-derivation.md
TalDerei: filed https://github.com/tachyon-zcash/tachyon/issues/79 for tracking.

crates/tachyon/src/stamp/proof/delegation.rs
TalDerei: yeah that's a fair observation.

book/src/nullifier-derivation.md
turbocrime: more speculatively: i think it might be useful to scale arity as the tree descends.

```mermaid
flowchart TB
  root ---> key0 & key1 & key2 & key3 & key4 & key5 & key6 & key7 & key8 & key9 & keyA & keyB & keyC & keyD & keyE & keyF
  key1 ---> key1.0 & key1.1 & key1.2 & key1.3 & key1.4 & key1.5 & key1.6 & key1.7
  key1.3 ---> key1.3.0 & key1.3.1 & key1.3.2 & key1.3.3
  key1.3.1 ---> key1.3.1.0 & key1.3.1.1
  key1.3.1.1 --> nf1.3.1.1
```

a structure like this would mean derivation of a large range is top-biased, which is an excellent tradeoff if we expect ranges to be relatively short.

turbocrime: for fixed arity and $a^d = 2^{16}$ worst-case edge $(a-1)d$ then arity 4 is optimal

crates/tachyon/src/stamp/proof/delegation.rs
TalDerei: addressed in https://github.com/tachyon-zcash/tachyon/pull/92

PR #63, opened 2026-04-28
Fix circular dependency with librustzcash

S1nus:
Although it's technically safe to depend on certain subcrates of librustzcash as long as those crates don't import tachyon, it smells of a circular dependency, and is easily avoidable in this case.

We can simply re-implement CompactSize and remove the dep.


TalDerei (commented):
well as you pointed out, the circular dependency is a latent risk; can we actually claim at this stage that the tachyon crate won't need to depend on `zcash_primitives` in any capacity going forward? also not a fan of vendoring deps like this, but don't have a better idea yet. cc @turbocrime

turbocrime:
we should bring this back in, but i'd like a more robust impl. will push some commits

turbocrime (approved):
roll these back if you'd like

Cargo.toml
```diff
@@ -135,6 +134,7 @@ rc_mutex = "forbid"
 
 # -- deny: restrictions (may need localized #[expect]) -----------------------
 absolute_paths = "deny"
+as_conversions = "deny"
```
turbocrime, line 137: loosening of `as_conversions` to permit these lines, which resist try-conversion due to const

https://github.com/S1nus/tachyon/blob/7fda8ccdef4315f0b009c2099a71955ab5347c68/crates/tachyon/src/serialization/compactsize.rs#L40-L47

i think this is acceptable after reading detail on this lint here https://rust-lang.github.io/rust-clippy/master/index.html?search=as_conversions#as_conversions

TalDerei (approved):

PR #64, opened 2026-04-28
Proof multiset improvements and hash chains

turbocrime:
Replaces the flat per-epoch pool polynomial with a single Poseidon anchor chain that advances one stamp at a time, and rebuilds the inclusion/exclusion proof primitives around per-stamp pool state.

`Anchor` advances through three domains:

- `Anchor::next_stamp(&TachygramSetCommit)` (`Tachyon-StampFld`) advances by one stamp
- `Anchor::next_empty()` (`Tachyon-EmptyBlk`) advances through a zero-stamp block, preserving per-height anchor uniqueness
- `Anchor::next_epoch(EpochIndex)` (`Tachyon-EpochStp`) advances across an epoch boundary

Each kind of advance carries its own Poseidon domain separator, so a segment built from one cannot pass for another.

- **`AnchorChain`** intra-epoch span used by `StampLift`. Built per-stamp by `AnchorSeed` / `EmptyBlockSeed`, composed by `AnchorFuse`.
- **`Unspent`** nullifier range-exclusion span used by `SpendableLift` and `SpendableEpochLift`. Built per-stamp by `UnspentSeed` / `EmptyBlockUnspentSeed`, composed by `UnspentFuse`.

`SpendableInit` bootstraps a `SpendableHeader` from the wallet's `NullifierHeader` and the tachygram set of the stamp that produced the note, advancing the anchor through that stamp. `SpendableLift` and `SpendableEpochLift` carry the spendable forward over `Unspent` spans, with `SpendableRollover` bridging epoch boundaries.

The GGM nullifier leaf key now takes a separate `Tachyon-NfDerive` step. Bundle commitment renamed `ZTxIdTachyonHash` and action digest renamed `Tachyon-ActionDg`.

New book chapters under `book/src/`: `proof-tree.md` (full PCD tree: headers, steps, lifecycle, roles, soundness, mermaid examples), `anchor.md`, `nullifiers.md`, `tachygrams.md`. `notes.md` and `domain-separators.md` updated to match.


book/src/anchor.md
```diff
@@ -0,0 +1,17 @@
...
+
+$$\mathsf{anchor}_h = \text{Poseidon}_\text{Tachyon-AnchrFld}(\mathsf{anchor}_{h-1}, h, \mathsf{block\_state}_h)$$
+
+## Block state
+
+The block-state input to the anchor is itself a running Poseidon hash, accumulating each stamp's tachygram set commitment in landing order:
+
+$$\mathsf{block\_state}_i = \text{Poseidon}_\text{Tachyon-PoolFold}(\mathsf{block\_state}_{i-1}, \mathsf{stamp\_commit}_x, \mathsf{stamp\_commit}_y)$$
+
+where $(\mathsf{stamp\_commit}_x, \mathsf{stamp\_commit}_y)$ are the affine coordinates of the $i$-th stamp's [tachygram set commitment](./tachygrams.md). The fold starts from a constant **block-state seed** — a single-input Poseidon hash of the `Tachyon-PoolFold` domain tag — which also stands as the closing block state of any empty block. The final $\mathsf{block\_state}$ is what consensus splices into the next anchor.
```
TalDerei: ACK – refreshing my memory for accumulator heirachy; at the tachystamp-level, we have a tachgram vector commitment (pedersen commitment to a multiset polynomial whose roots are poseidon hashes). at the block-level, we have a block accumulator (poseidon hash-chain that aborbes the coordinates of each stamps pedersen commitments).

crates/tachyon/src/stamp/proof/mod.rs
TalDerei, line 1: meta commentary to trace my understanding; the crux of this PR replaces the per-epoch pool polynomial that existed with a continuous poseidon hash chain (and the anchor associatively becomes a field element). now there's other details i'm abstracting, but the rough flow is:
- `SpendableInit` (inclusion proof that cm is in the stamp) is basically a bootstrap step here for a wallet reengaging with the OSS after say a wallet comes back online after being offline for some time,
- it witnesses one stamp (the one containing commitment), not the whole epoch pool which was the previous status quo,
- it outputs a `SpendableHeader`,
- then `SpendableLift` updates the spendable forward to the current anchor (internally consumes `Unspent` shards, which are exclusion proofs that the nullifier is not in the stamp), 

so both `SpendableInit` / `SpendableLift` steps both used to witness the whole epoch's polynomial, but now only witness a single stamp's tachygram set. this implies to me that the wallet cost has in some sense a O(1) complexity cost? it needs to:
- `NfMasterSeed` + descend the GGM tree (pre-blind step): maybe ~D SNARKs
- `SpendableInit`: 1 SNARK (witnesses one stamp)

so the wallet's now proving things about it's own stamps not entire epoch pool. the OSS on the other hand still has this idealized O(epoch history) cost.

crates/tachyon/src/stamp/proof/pool.rs
```diff
@@ -0,0 +1,388 @@
+//! Chain-bound primitives over consensus block state.
+//!
+//! Hosts the within-block tg-bound shards (`InclusionShard`,
+//! `ExclusionShard`), the within-block tg-unbound rollback chain
+//! (`InclusionComplement`), and the raw multi-block span (`AnchorSpan`).
+//! The tg-bound multi-block span (`Unspent`) and its single-block seeds
+//! live in the sibling `unspent` module.
+//!
+//! See the plan at `73a4847-implemented-a-new-snoopy-crown.md` for the
+//! design rationale and step semantics.
```
TalDerei: remove

crates/tachyon/src/digest/poseidon.rs
```diff
@@ -14,9 +14,9 @@ const NOTE_MASTER_DOMAIN: &[u8; 16] = b"Tachyon-MkDerive";
...
 const NOTE_COMMITMENT_DOMAIN: &[u8; 16] = b"Tachyon-CmDerive";
 const DELEGATION_DOMAIN: &[u8; 16] = b"Tachyon-Delegate";
-const SUB_BLOCK_DOMAIN: &[u8; 16] = b"Tachyon-StampFld";
-const ANCHOR_BLOCK_DOMAIN: &[u8; 16] = b"Tachyon-BlockFld";
-const EPOCH_INCREMENT_DOMAIN: &[u8; 16] = b"Tachyon-EpochInc";
+const STAMP_FOLD_DOMAIN: &[u8; 16] = b"Tachyon-StampFld";
+const EMPTY_BLOCK_DOMAIN: &[u8; 16] = b"Tachyon-EmptyBlk";
+const EPOCH_STEP_DOMAIN: &[u8; 16] = b"Tachyon-EpochStp";
 const ACTION_DOMAIN: &[u8; 16] = b"Tachyon-ActnDgst";
 const PAYMENT_KEY_DOMAIN: &[u8; 16] = b"Tachyon-PkDerive";
```
TalDerei: can we document all of these domain seperators in code? there also appears to be some drift against the `anchors.md` and `domain-seperators.md` book entries.

crates/tachyon/src/primitives/anchor.rs
```diff
@@ -1,68 +1,164 @@
...
+    ///
+    /// Panics if `stamp_commit` is the identity point.
+    #[must_use]
+    pub fn next_stamp(self, stamp_commit: &TachygramSetCommit) -> Self {
+        let point = stamp_commit.0.inner();
+        let coords = point
+            .coordinates()
+            .expect("must not be identity commitment"); // TODO: error?
+        Self(poseidon::anchor_stamp_step(self.0, coords))
+    }
```
TalDerei, line 36: this panic is computational infeasible, but would return a `Result` for hygiene.

crates/tachyon/src/stamp/proof/spendable.rs
```diff
@@ -89,157 +132,207 @@ impl Step for SpendableInit {
...
 
-    const INDEX: Index = Index::new(6);
+    const INDEX: Index = Index::new(13);
 
     fn witness<'source>(
         &self,
-        (pool, anchor): Self::Witness<'source>,
-        (cm_tg, nf, epoch): <Self::Left as Header>::Data<'source>,
+        (pre_cm_anchor, stamp_tg_set): Self::Witness<'source>,
+        (cm, nf, _epoch): <Self::Left as Header>::Data<'source>,
```
TalDerei, line 169: why does `SpendableInit` drop `_epoch` from `NullifierHeader`?

TalDerei (commented):
first pass observations

book/src/anchor.md
turbocrime: sorry, i folded in #75 and this is out of date now

crates/tachyon/src/stamp/proof/mod.rs
turbocrime: the user proves inclusion in stamps that landed on chain. those might be its autonomes, but they're probably aggregated.

envelope math with present block size and expected stamp size is, an ideally aggregated block with one stamp might have ~5k items. that provides upper bound for the worst-case poly claim and it might become fairly common.

book/src/anchor.md
TalDerei: sorry should've been explicit; I know the docs are stale after folding in #75, this traces my understanding of the accumulators in the current code.

crates/tachyon/src/stamp/proof/spendable.rs
turbocrime: the nullifier binds the epoch, and now everything downstream relies on that.

crates/tachyon/src/stamp/proof/mod.rs
TalDerei: sure stamp's polynomial could represent an autonome or aggregated, but the mental model here is directionally correct?

crates/tachyon/src/stamp/proof/spendable.rs
turbocrime: also, it is no longer possible to open the anchor commitment to check the block height. we simply rely on collision resistance to provide continuity later.

turbocrime: epoch is still on the input because it's used in another location to confirm epoch rollover

crates/tachyon/src/primitives/anchor.rs
turbocrime: imo this is a possible DoS vector when deserializing and then inspecting adversarial input, so yes.

but prev discussion on this kind of thing pointed towards "just panic"

https://github.com/tachyon-zcash/tachyon/pull/34#discussion_r2920370205

we should pick an attitude and be consistent.

crates/tachyon/src/stamp/proof/mod.rs
turbocrime: yes.

inclusion test just cm+nf derivation, a set test, and anchor step.

need to privately lift to block, which is one set test and anchor step per following stamp.

no more large poly queries. biggest poly we handle is one stamp tg set.

if set tests are not very expensive, then ggm is the heaviest part of the process.

TalDerei: yeah this traces with my mental model, great!

crates/tachyon/src/primitives/anchor.rs
TalDerei: can we file an issue that standardizes our error handling postering? don't necessarily need to block on this

TalDerei: leaning towards `Result` because that seems more idiomatic error handling.

crates/tachyon/src/stamp/proof/mod.rs
TalDerei: don't have a good intuition which would be more expensive, but they both seem relatively inexpensive.

turbocrime: iirc, additional proof tree complexity consumes things like wire capacity and increases proving cost?

so if they're roughly equal, the simpler tree is preferable

also, just fewer moving parts.

crates/tachyon/src/primitives/anchor.rs
TalDerei: filed https://github.com/tachyon-zcash/tachyon/issues/82

crates/tachyon/src/bundle/mod.rs
TalDerei, line 1: `anchorTachyon` is still quoted as 64 bytes, we should call it the "running anchor (Fp)" and change it to 32 bytes.

crates/tachyon/src/stamp/proof/stamp.rs
```diff
@@ -67,7 +63,7 @@ impl Step for OutputStamp {
         Anchor,
```
TalDerei, line 72: it's the case that the anchor here is freely witnessed (with no in-circuit binding), so consensus is responsible for verifying it's a real published anchor before accepting the stamp. it'd be nice to maybe have some kind of book entry on things consensus needs to enforce like this.

TalDerei (approved):
optimistically to avoid blocking, we can iterate. assuming we fix the discussed book drift and inconsistent error handling.

book/src/anchor.md
turbocrime: outright removed the worst docs that need a rewrite, updated some others

crates/tachyon/src/digest/poseidon.rs
turbocrime: these are all documented in the book now.

constants are now located closer to their relevant digest functions to improve understanding of relationships.

crates/tachyon/src/stamp/proof/stamp.rs
turbocrime: it's freely witnessed, but can't merge with any spends unless it's a real member of the chain.

will add some comments wherever an anchor is freely witnessed to describe future step binding or consensus requirements.

crates/tachyon/src/bundle/mod.rs
turbocrime: made a pass to cleaning up serialization drift

crates/tachyon/src/stamp/proof/stamp.rs
turbocrime: comprehensively reviewed header and step types to annotate with provenance and soundness information

book/src/anchor.md
turbocrime: rewrote a few chapters

PR #65, opened 2026-04-30
use no-std friendly io::Error constructor

S1nus:
The current method of constructing `io::Error`s is not no_std friendly, here's a fix for that.


TalDerei:
what's the status of this?

turbocrime (approved):
rebased

PR #70, opened 2026-05-01
test: add coverage for keys, action, and effect modules

LucidSamuel:


TalDerei:
this might be premature since there's a lot of inflight PRs that change things, i'd make this a draft and we can iterate on test coverage accordingly. cc @turbocrime

turbocrime (commented):
i updated this since it was simple but then noticed #71 has some overlap

i think most of these would also benefit from the kind of detail provided by proptest, so it might make sense to fold them together

PR #71, opened 2026-05-03
feat: property-based testing harness for tachyon types

### LINKED ISSUE #18: Property-based testing
`proptest = "1"` is already in `Cargo.toml` as an unconditional dependency but is completely unused. No generators, no property tests.

## Current state

- `proptest = "1"` in `crates/tachyon/Cargo.toml:30` — unconditional (should be `[dev-dependencies]`)
- No `proptest` generators or `Arbitrary` implementations anywhere
- No `pub mod testing` on any type
- All existing tests use deterministic seeds: `StdRng::seed_from_u64(0)` etc. in bundle.rs, action.rs, value.rs, keys/mod.rs, entropy.rs

## Needed generators

- `arb_spending_key() -> SpendingKey` — random 32-byte arrays
- `arb_note() -> Note` — valid (pk, value, psi, rcm) combinations
- `arb_value() -> Value` — u64 bounded by `NOTE_VALUE_MAX`
- `arb_commitment_trapdoor() -> CommitmentTrapdoor` — random Fp elements
- `arb_nullifier_trapdoor() -> NullifierTrapdoor` — random Fp elements
- `arb_action_plan() -> action::Plan` — spend/output with valid notes and entropy
- `arb_bundle_plan() -> BundlePlan` — balanced sets of actions

## Key properties to test

- **Round-trip serialization**: for every type with `Into<[u8; N]>` / `TryFrom<&[u8; N]>`, encoding then decoding yields the original value
- **Homomorphic commitment**: `commit(v1, r1) + commit(v2, r2) == commit(v1+v2, r1+r2)` — the `Sum` impl on `value::Commitment` should preserve this
- **Signature verify-after-sign**: sign with any valid (ask, alpha, sighash), verify with the corresponding rk — must always pass
- **Balance conservation**: for a balanced bundle plan, `sum(spend_values) - sum(output_values) == value_balance`
- **Action digest determinism**: same action inputs always produce the same `ActionDigest`
- **Key derivation consistency**: `SpendingKey -> SpendAuthorizingKey -> AuthorizationKey` is deterministic; same sk always yields same ak
- **Value bounds**: `Value::from(n)` where `n > NOTE_VALUE_MAX` should panic (or be rejected if bounds checking changes to `TryFrom`)

## Orchard reference

Every public type has `pub mod testing` with `proptest` generators: `arb_spending_key()`, `arb_note()`, `arb_bundle()`, `arb_action()`, `arb_note_value()`, etc.

LucidSamuel:
fixes #18

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 4
 
 [[package]]
+name = "anyhow"
+version = "1.0.102"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
+
+[[package]]
 name = "arrayref"
 version = "0.3.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


TalDerei:
same feedback as https://github.com/tachyon-zcash/tachyon/pull/70#issuecomment-4467871233, cc @turbocrime.

PR #74, opened 2026-05-11
various refactors

turbocrime:
Pulled out of #64 to reduce diff. Stacks on #62.

- `bundle.rs` split to `bundle/{mod.rs, tests.rs}`
- `test_support.rs` renamed `fixtures.rs`.
- new `digest` module organizes utils and constants
- newtype `Fp` conversions flip to owned form
- minor mock api change
- mock error type carries a message


crates/tachyon/src/digest/blake2b.rs
```diff
@@ -0,0 +1,88 @@
...
+use blake2b_simd::Params;
+
+const PRF_EXPAND_PERSONALIZATION: &[u8; 16] = b"Zcash_ExpandSeed";
+pub(crate) const BUNDLE_COMMITMENT_PERSONALIZATION: &[u8; 16] = b"Tachyon-BndlHash";
+pub(crate) const AUTH_DIGEST_PERSONALIZATION: &[u8; 16] = b"ZTxAuthTachyHash";
+const SPEND_ALPHA_PERSONALIZATION: &[u8; 13] = b"Tachyon-Spend";
+const OUTPUT_ALPHA_PERSONALIZATION: &[u8; 14] = b"Tachyon-Output";
+
+// Domain separators 0x00–0x05 are Sapling, 0x06–0x08 are Orchard.
+// Tachyon allocates 0x09+ to avoid collisions.
```
TalDerei: can we investigate the veracity of this claim? i'd also like to model this file more closely after https://github.com/zcash/zcash_spec/blob/main/src/prf_expand.rs, but this can be done as follow-up work.

TalDerei (approved):
ACK pure structural refactor.

crates/tachyon/src/digest/blake2b.rs
turbocrime: lol this is so wrong. thanks

turbocrime: relocated to 0x20 range

PR #75, opened 2026-05-14
Single hash chain

turbocrime:
Stacks on #64.

Collapses the per-block hash chain from #64 into a single anchor
sequence that advances one stamp at a time. Drops `SubBlock`,
`InclusionShard`, `InclusionComplement`, and the entire `unspent.rs`
module; trims the proof system to one running anchor with three
domain-separated link types.

## Anchor links

Three domain-separated `Anchor` advances replace #64's
`next_block` / `next_epoch` pair plus the `SubBlock` per-block hash:

- `Anchor::next_stamp(&TachygramSetCommit)` (`Tachyon-StampFld`) absorbs
  one stamp's tachygram-set commit.
- `Anchor::next_empty()` (`Tachyon-EmptyBlk`) advances through a
  zero-stamp block, preserving per-height anchor uniqueness.
- `Anchor::next_epoch(EpochIndex)` (`Tachyon-EpochStp`) lifts across an
  epoch boundary; emitted only by `SpendableRollover`.

Block alignment is now a consensus convention — validators check that
anchor endpoints match the published per-block anchor sequence.
[crates/tachyon/src/primitives/anchor.rs](crates/tachyon/src/primitives/anchor.rs).

## Pool primitives, single-level

[crates/tachyon/src/stamp/proof/pool.rs](crates/tachyon/src/stamp/proof/pool.rs)
now hosts both the nf-free anchor segments and the per-nf exclusion
chains, all seeded per stamp:

- **`AnchorChain`** (`(start, end)`) — nf-free anchor segment used by
  `StampLift`. Built per-stamp by `AnchorSeed`, per-empty-block by
  `EmptyBlockSeed`, and composed by `AnchorFuse`. Replaces #64's
  `AnchorSpan`.
- **`Unspent`** (`(nf, start, end)`) — per-nf range-exclusion proof.
  Built per-stamp by `UnspentSeed` (verifies `nf ∉ stamp_tg_set` and
  absorbs the stamp), per-empty-block by `EmptyBlockUnspentSeed`, and
  composed by `UnspentFuse`. Multi-block fusion now spans block
  boundaries directly because anchor advances are continuous.

Deleted: `InclusionShard` / `InclusionShardFuse`, `ExclusionShard` and
its seed/fuse, `InclusionComplement` and its seed/step. The entire
[crates/tachyon/src/stamp/proof/unspent.rs](crates/tachyon/src/stamp/proof/unspent.rs)
module is gone.

## Simpler spendable bootstrap

`SpendableInit` now fuses the wallet's pre-blind `NullifierHeader{cm,
nf, epoch}` directly with a freely-witnessed `(pre_cm_anchor,
stamp_tg_set)`. It verifies `cm ∈ stamp_tg_set` and emits
`SpendableHeader{nf, pre_cm_anchor.next_stamp(commit)}` — the running
anchor immediately after the cm-stamp's absorption. No `InclusionShard`
construction, no `InclusionComplement` rollback, no chain matching.

`pre_cm_anchor` is freely witnessed; the spendable reaches a real
published anchor only by extending through `SpendableLift` over
`Unspent` segments built from real stamps. The cm-block itself cannot
contain the wallet's `nf` (`nf` is GGM-bound to `cm`), so no
nf-exclusion check is needed at the cm-stamp.

`SpendableLift` / `SpendableEpochLift` / `StampLift` and
`SpendableRollover` keep their #64 semantics — only their inputs'
internal shape changed.


turbocrime:
folding into #64

Issue #84, opened 2026-05-19
Close block state with metadata to produce an anchor


turbocrime:
in #64 the anchor is a continuous hash chain over stamps, with sentinel steps for empty blocks and epoch boundaries.

similar to the epoch boundary step, we could include a domain-separated block boundary. such a step may commit block metadata into the anchor.

such a refactor this may or may not involve a distinct header and/or distinct primitive sub-block states.

consider that many proof steps freely witness an anchor that is bound later.

PR #91, opened 2026-05-20
introduce `CloseBlockSeed` and `CloseBlockUnspentSeed`

turbocrime:
Exploration of #84.

Adds a domain-separated block-closing step to the anchor hash chain that absorbs `(prev_anchor, epoch_index)`. Replaces the empty-block sentinel: every block, empty or not, now ends with one `close_block` after any per-stamp absorbs. Per-height uniqueness already follows from chain continuity, so the only metadata the close step needs to commit is the block's epoch.

- `digest::poseidon`: `anchor_empty_step` → `anchor_close_step(prev, epoch_index)`, domain `Tachyon-BlockEnd`.
- `primitives::anchor`: `Anchor::next_empty` → `Anchor::close_block(epoch: EpochIndex)`. `Anchor::next_epoch` is unchanged.
- `stamp::proof::pool`: `EmptyBlockSeed` → `CloseBlockSeed` (witness `(start, epoch)`); `EmptyBlockUnspentSeed` → `CloseBlockUnspentSeed` (witness `(start, epoch, nf)`). PCD INDEXes retained. Output headers (`AnchorChain`, `Unspent`) unchanged.
- `stamp::proof::spendable`: unchanged. `SpendableRollover` still applies `Anchor::next_epoch` to produce `boundary_anchor` and `SpendableEpochLift` still consumes it — `next_epoch` remains the only cross-epoch gate.
- `fixtures`: `anchor_at`, `build_anchor_chain_pcd`, `build_unspent_pcd`, and `spendable_init` emit one close seed per block. `spendable_init`'s "skip lift when cm is last stamp" early return is gone; the close step is now always part of the cm-block's lift.

The proof tree handles anchors freely so didn't change much.

`SpendableInit` still witnesses at a sub-block state. `SpendableLift`, `SpendStamp`, `MergeStamp`, `StampLift`, and `OutputStamp` all carry and equality-check anchors without inspecting kind.

The usage pattern this implies is the one wallets and aggregators already follow: a spendable starts intra-block at `SpendableInit` and should be privately lifted, just now with one more fuse and `CloseBlockUnspentSeed`.

The logic of `Unspent` ranges didn't change, but likewise a `CloseBlockUnspentSeed` is necessary per block. Same for `AnchorChain` which gains a `CloseBlockSeed` per block.

## todo

- **Book updates.** `book/src/anchor.md`, `book/src/proof-tree.md`, `book/src/domain-separators.md` still describe the empty-block sentinel.
- **Primitive type split.** Possibly bring back `SubBlock` primitive.
- **Proof headers and steps.** Possibly restore distinct sub-block header types and restrict range proofs to full-block inputs only.
- **Proof logic improvements.** Once block boundaries commit to epoch, the proof tree may wish to open these commitments at consumers to constrain epoch directly — possibly subsuming `next_epoch` and the boundary-anchor gating at `SpendableRollover`/`SpendableEpochLift`.

### DIFF
--- a/crates/tachyon/src/digest/poseidon.rs
+++ b/crates/tachyon/src/digest/poseidon.rs
@@ -119,14 +119,18 @@ pub(crate) fn anchor_stamp_step(anchor_prev: Fp, tgs: Coordinates<EqAffine>) -> 
     ])
 }
 
-const ANCHOR_EMPTY_DOMAIN: &[u8; 16] = b"Tachyon-EmptyBlk";
+const ANCHOR_CLOSE_DOMAIN: &[u8; 16] = b"Tachyon-BlockEnd";
 
-/// Advances the anchor through one block that contains zero stamps.
+/// Closes a block: absorbs the block's epoch index into the running anchor.
+///
+/// Runs exactly once per block at end-of-block, after any stamp absorbs.
+/// A block with zero stamps still runs this step on its starting anchor.
 #[must_use]
-pub(crate) fn anchor_empty_step(anchor_prev: Fp) -> Fp {
-    hash::<2>([
-        Fp::from_u128(u128::from_le_bytes(*ANCHOR_EMPTY_DOMAIN)),
+pub(crate) fn anchor_close_step(anchor_prev: Fp, epoch_index: u32) -> Fp {
+    hash::<3>([
+        Fp::from_u128(u128::from_le_bytes(*ANCHOR_CLOSE_DOMAIN)),
         anchor_prev,
+        Fp::from(u64::from(epoch_index)),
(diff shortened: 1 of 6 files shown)


turbocrime (commented):
this is a very minimal approach, and i don't like it.

we should consider getting rid of the epoch sentinel.

we should probably have a 'sub-block state' header representation, to scope that structurally. closing a sub-block state would become a single step, instead of a seed-and-fuse. also, we could reorient range proofs to the block boundaries that become evident.

turbocrime:
closing in favor of https://github.com/tachyon-zcash/tachyon/pull/117

PR #92, opened 2026-05-20
Mock Ragu: add index enforcement

turbocrime:
Brings real ragu's three-layer step/header enforcement to the mock,
which previously did none of it.

- **Register-time**: `ApplicationBuilder::register` requires sequential `Index::new(0..)` and rejects two distinct `Header` types declaring the same `SUFFIX` (deduped by `TypeId`).
- **Header hash**: `compute_header_hash` mixes in `SUFFIX`, so a proof bound to Header A can't be verified as Header B.
- **Verify**: `Proof` now carries `step_index`; verify rejects any step_index outside the registered application range.

`Application::rerandomize` intentionally stays a transformation (not a fusion via an internal Rerandomize step) since the mock has no recursion; the API surface is unchanged.

Tests cover both register-time and verify-time enforcement.

Tachyon's 24-step registration is already sequential with no SUFFIX collisions, so it passes through unchanged.

### DIFF
--- a/crates/mock_ragu/src/header.rs
+++ b/crates/mock_ragu/src/header.rs
@@ -4,10 +4,11 @@ use alloc::vec::Vec;
 
 /// Number of internal header suffixes reserved by mock_ragu.
 ///
-/// Only the trivial header [`()`] uses an internal suffix; reserving a small
-/// range mirrors real ragu's value-space partition between internal and
-/// application suffixes.
-pub(crate) const NUM_INTERNAL_SUFFIXES: usize = 1;
+/// Mirrors real ragu's `InternalStepIndex` layout:
+/// - Slot 0: `Rerandomize` (reserved; mock rerandomize is a transformation,
+///   not a Step, but the slot stays reserved for migration parity).
+/// - Slot 1: trivial header [`()`].
+pub(crate) const NUM_INTERNAL_SUFFIXES: usize = 2;
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd)]
 enum HeaderSuffix {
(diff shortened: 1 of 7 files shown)


crates/mock_ragu/src/header.rs
```diff
@@ -2,15 +2,61 @@
 
 use alloc::vec::Vec;
 
+/// Number of internal header suffixes reserved by mock_ragu.
+///
+/// Only the trivial header [`()`] uses an internal suffix; reserving a small
+/// range mirrors real ragu's value-space partition between internal and
+/// application suffixes.
```
TalDerei: ACK – sufficiently addresses suffix collision in https://github.com/tachyon-zcash/tachyon/pull/62#discussion_r3251491683.

crates/mock_ragu/src/header.rs
```diff
@@ -2,15 +2,61 @@
...
+pub(crate) const NUM_INTERNAL_SUFFIXES: usize = 1;
```
TalDerei: `NUM_INTERNAL_SUFFIXES` and `NUM_INTERNAL_STEPS` constants diverge from ragu's slot layout; maybe we should just target parity.

TalDerei (approved):
seems like a faithful port, and we can iterate on stricter parity as mock ragu moves into the empty [ragu crate](https://github.com/tachyon-zcash/ragu/blob/main/src/lib.rs).

crates/mock_ragu/src/header.rs
turbocrime: this is due to internal details, such as the mock 'rerandomize' is not actually a step.

i think a more 'correct' implementation would be awkward and also not very useful for mock purposes.

TalDerei: I know why there's a divergence, i'm just noting that mock ragu will need to target explicit parity with our existing API, including these constants. i'm suggesting it should be a step.

TalDerei: some OOB context, Sean is looking for API parity explicitly as mock ragu eventually migrates to https://github.com/tachyon-zcash/ragu/tree/main/src, so i'm anticipating this being a point of divergence that will need to be corrected eventually.

turbocrime: i'll expand the reserved indexes

Issue #99, opened 2026-05-23
Provide linear range extension steps


turbocrime:
Presently, range proofs of any kind are established by seeding each individual point in a range, and then performing binary fuse. This provides `O(2N-1)` step complexity for any range of size N and may be parallelized efficiently. But technically, a linear range *extension* would only be `O(N)` step complexity.

Consider that any given hardware has practical limits to *how parallel* its tasks may be, usually related to core count. So in some contexts, it may be preferable to perform range extension rather than a range fuse. We may conservatively assume that a sufficiently large range will pass this inflection point for any hardware.

So, every range-proof subtree should provide both binary-fuse steps and transform-extend steps to provide composable utilities that consumers may apply to schedule ideal range proofs in their use-case.

This may be combined with packing approaches informed by https://github.com/tachyon-zcash/ragu/pull/720

Issue #98, opened 2026-05-23
Explore block-summary or range-summary proofs


turbocrime:
Since #64, inclusion and exclusion tests against individual stamps. But for many purposes, a block is the natural unit size.

Since it's expected to be possible for a *single massive stamp* to occupy an entire block, we can assume that any less-efficient use of block space is also possible to represent as a *single less massive stamp*.

Thus, if any block lands containing multiple stamps, it should be feasible to produce a 'block summary' proof carrying all block tachygrams and a set of the intermediate states within a block, and use this as a foundation for any inclusion or exclusion tests involving the block.

Alternatively, this could be a generic range-summary proof, which doesn't bother to distinguish between block-end anchors and sub-block states, which would fit naturally unless/until #84 lands.

This may conceivably replace the present range proofs `AnchorChain` etc.

PR #117, opened 2026-05-25
Introduce `RangeSummary` proof

turbocrime:
Work on #98

Introduces `RangeSummary`, a bounded anchor segment that additionally carries the union of every absorbed stamp's tachygram set, and reworks the spendable bootstrap and per-nf exclusion path to consume it.

A `RangeSummary` is assembled via `RangeSummaryStampSeed` / `RangeSummaryEmptySeed` / `RangeSummaryFuse`. It may be feasible to add more elaborate steps that improve composition https://github.com/tachyon-zcash/tachyon/issues/99. A range may grow only as long as the multiset content stays under a defined maximum (presently 8192 items).

A `SpendableInitRange` fuses the wallet's `NullifierHeader` with a `RangeSummary` covering the inclusion stamp and verifies both `cm ∈ range.tg_set` (inclusion) and `nf ∉ range.tg_set` (exclusion). The exclusion check is new and required: because the bootstrap range can cover many stamps, the nullifier could have already been spent within it, so the bootstrap now checks exclusion the same way every downstream `SpendableLift` does.

`Unspent` is now also produced via `UnspentRange`, which consumes a `RangeSummary`, witnesses a nullifier + matching tachygram set, and emits an `Unspent` after proving `nf ∉ range.tg_set`.

The range additions are purely additive: every original path is retained. `SpendableInitStamp` still bootstraps a spendable directly from an individual stamp, and `UnspentSeed` / `EmptyBlockUnspentSeed` still seed an `Unspent` from a single stamp or empty block without an intervening `RangeSummary`. `UnspentRange` is exactly equivalent to a `RangeSummaryStampSeed` followed by `UnspentSeed`, so it adds no new trust surface, only a reusable intermediate.

This should improve sync service costs for providing `Unspent` proofs. The `RangeSummary` should be re-useable input for any nullifier exclusion test, and it might also be served to any user device interested in initializing a spendable.

### DIFF
--- a/book/src/proof-tree.md
+++ b/book/src/proof-tree.md
@@ -17,7 +17,7 @@ A note is essentially created when the wallet runs `SpendableInitStamp` or `Spen
 
 Maintaining the spendable means advancing its anchor forward over `Unspent` segments.
 `SpendableLift` consumes one `Unspent` whose nullifier matches the spendable's and whose start matches the spendable's current anchor, producing a fresh `SpendableHeader` at the segment's end.
-Each `Unspent` is built either by running `UnspentFromRange` against a `RangeSummary` or directly from a single stamp via `UnspentSeed` / `EmptyBlockUnspentSeed`; `UnspentFuse` composes adjacent segments.
+Each `Unspent` is built either by running `UnspentRange` against a `RangeSummary` or directly from a single stamp via `UnspentSeed` / `EmptyBlockUnspentSeed`; `UnspentFuse` composes adjacent segments.
 
 Crossing an epoch boundary requires a new-epoch nullifier and the cross-epoch anchor advance.
 `RolloverFuse` consumes two consecutive-epoch `NullifierHeader`s that share a note commitment and emits a `NullifierRolloverHeader`.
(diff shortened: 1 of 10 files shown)


crates/tachyon/src/stamp/proof/pool.rs
```diff
@@ -1,36 +1,47 @@
...
+//! Hosts two range shapes:
+//!
+//! - [`AnchorChain`] — set-free anchor continuity, consumed by
+//!   [`super::stamp::StampLift`] to advance a stamp's anchor.
+//! - [`RangeSummary`] — bounded anchor segment that additionally carries the
+//!   union of every absorbed stamp's tachygram set, the foundation for
+//!   inclusion and exclusion tests against a segment.
+//!
+//! Per-nf exclusion lives in [`super::spendable`] as `Unspent`, built
+//! from a [`RangeSummary`] via `UnspentFromRange`.
```
TalDerei: nit; 'union' of of tachygram sets might be a slight misnomer here? tracing how merge works, it double counts shared elements since it's a multiset, so maybe “merged multiset” would be more precise?

crates/tachyon/src/stamp/proof/spendable.rs
```diff
@@ -34,35 +34,156 @@ use pasta_curves::Fp;
...
-    /// in [`super::delegation::NullifierHeader`]. `anchor` is computed
-    /// at [`SpendableInit`] as `pre_cm_anchor.next_stamp(stamp_commit)`
-    /// over a freely-witnessed `pre_cm_anchor`, then advanced over
-    /// [`Unspent`] segments at [`SpendableLift`] /
-    /// [`SpendableEpochLift`].
+    /// in [`super::delegation::NullifierHeader`]. `anchor` is inherited
+    /// at [`SpendableInit`] as the `end` of a [`RangeSummary`] whose
+    /// `tg_set` contains the note's commitment and excludes the
+    /// nullifier, then advanced over [`Unspent`] segments at
+    /// [`SpendableLift`] / [`SpendableEpochLift`].
```
TalDerei, line 267: nit; can we add positive / negative multi stamp `RangeSummary` tests for `SpendableInit`?

book/src/proof-tree.md
```diff
@@ -326,6 +335,7 @@ A sync-service variant substitutes `DelegateRolloverFuse` (consuming two `Delega
 | Header | Fields |
 | ------ | ------ |
 | AnchorChain | (prev_anchor, end_anchor) |
+| RangeSummary | (start, end, tg_set) |
```
TalDerei, line 358: ACK – `RangeSummary` witnesses start anchor + each stamps tachygram set, advances anchor and does a set union (merge stamp sets), and produces (start, end, `tg_set` commitment). 

in terms of improving sync costs, my rough read is that before there was an `Unspent` step / nullifier, so M stamps for N nullifiers = N * M proofs. now, the per stamp work is agnostic to nf, so we can build the `RangeSummary` once over M stamps and nullifier can run one step / query against whole range's set = M + N proofs.

book/src/proof-tree.md
```diff
@@ -357,7 +369,7 @@ A sync-service variant substitutes `DelegateRolloverFuse` (consuming two `Delega
 | DelegateNullifierStep | DelegateNfPrefixHeader | — | — | DelegateNullifierHeader |
 | RolloverFuse | NullifierHeader | NullifierHeader | — | NullifierRolloverHeader |
 | DelegateRolloverFuse | DelegateNullifierHeader | DelegateNullifierHeader | — | NullifierRolloverHeader |
-| SpendableInit | NullifierHeader | — | anchor, stamp_tg_set | SpendableHeader |
+| SpendableInit | NullifierHeader | RangeSummary | tg_gadget | SpendableHeader |
```
TalDerei: ACK – before, `SpendableInit` took one specific stamp's tachygram set and checked whether cm was in the stamp (stamp inclusion test), but now `SpendableInit` consumes `RangeSummary` where `tg_set` is union over many stamps, so it needs to do both a cm inclusion and nf exclusion test `UnspentFromRange` (wallet does both tests). one extra query seems marginally fine to me.

TalDerei (approved):
some minor nits; logic seems to check out to me against the proof tree, this is a clever design 👍

turbocrime:
closing in favor of #188 which is essentially the same, but more recent

PR #124, opened 2026-05-27
mock_ragu: replace with upstream mock

turbocrime:
Replaces the in-tree `mock_ragu` placeholder with the ragu mock module, and ports tachyon's primitives and PCD steps onto ragu's mock API.


crates/mock_ragu/src/gadgets.rs
```diff
@@ -50,3 +48,110 @@ impl Multiset {
...
+    /// Pass `Fp::ZERO` for a deterministic, untrapdoored commit. Otherwise
+    /// the trapdoor pins the commit to a caller-chosen value: `commit(t) ==
+    /// commit(t')` iff the polynomials match *and* `t == t'` (Pedersen
+    /// independence of the coefficient generators from the blinding
+    /// generator). Callers that need to pin a commit to an external identifier
+    /// supply that identifier as `trap`.
+    #[must_use]
+    pub fn commit(&self, trap: Fp) -> Commitment {
+        self.0.commit(trap)
+    }
```
TalDerei: doc comment makes a false claim and this seems like a bug; afaict trailing zeros are ignored here, so sets will have identical commits. the idea is although we never need to store a zero, we can't ignore them because the  threat model is adversarial and the design should probably be length bound anyways.

suggest using dedicated length generator, 

```ts
  pub fn commit(&self, trap: Fp) -> Commitment {
      let len = Fp::from(self.0.coefficients().len() as u64);
      self.0.commit(trap) + Commitment::sequence_length_generator() * len
  }
```

and then in new generator in polynomial file,

```ts
  static ref SEQUENCE_LENGTH_GENERATOR: EqAffine = {
      use pasta_curves::{arithmetic::CurveExt as _, group::Curve as _};
      Eq::hash_to_curve("mock_ragu:sequence_length")(b"len").to_affine()
  };

  pub fn sequence_length_generator() -> Self {
      Self(*SEQUENCE_LENGTH_GENERATOR)
  }
```

and `Sequence::PartialEq` can then compare coefficients directly,

```ts
  fn eq(&self, other: &Self) -> bool {
      self.0.coefficients() == other.0.coefficients()
  }
```

and then also some regression test.

TalDerei:
left an observation that needs correcting, otherwise good with having unordered multiset and ordered sequence poly concepts (that use the same `Polynomial` / `Commitment` under the hood).

turbocrime:
i think i'm actually going to reorient to use a bare polynomial and point, since that seems to be the more accurate approach.

crates/mock_ragu/src/gadgets.rs
turbocrime: does binding the length make sense? it doesn't seem like ragu actually does that, and anything i can implement in a consumer of ragu can't soundly bind the length.

TalDerei: doesn't matter for polynomial semantics in ragu, it's fine and trailing zeros yield the same poly. the motivation for length binding here is that tachyon needs sequence length to matter, so i'm suggesting we don't touch the PCS layer and just model this as like "extra protocol state" in some layer above. we shouldn't need to modify the raw poly commit semantics. these aren't mutually exclusive. 

maybe something like 

```ts
  struct SeqCommit {
      poly_commit: Commitment,
      len: usize,
  }
```

turbocrime: trying to reason about where length actually matters right now

most polys should either:
- be secret and maximum degree, in a way that only harms yourself if you violate
- be homomorphically related to a poly in the above class, so commitments can confirm

so afaict only the nullifier pair at spendbind is actually exposed to soundness risk. working on that now

TalDerei (commented):
suggesting putting this on hold until https://github.com/tachyon-zcash/ragu/pull/780 lands. then we can consume mock from the upstream ragu crate dep.

TalDerei:
can we collapse the mock diff here? any other changes to the mock should be upstreamed to ragu.

TalDerei:
> can we collapse the mock diff here? any further changes to the mock should be upstreamed to ragu.

it's also feature gated now.

crates/tachyon/src/stamp/proof/stamp.rs
```diff
@@ -157,47 +161,61 @@ impl Step for MergeStamp {
...
+            merged_poly
+        };
+
+        let merged_tachygram_poly = {
+            let left_poly = Polynomial::from(left_tachygram);
+            let right_poly = Polynomial::from(right_tachygram);
+            let merged_poly = left_poly.multiply(&right_poly);
+            enforce_poly_product(ctx, &left_poly, &right_poly, &merged_poly)?;
+            merged_poly
+        };
```
turbocrime, line 214: only actual use of poly tools from the mock

TalDerei (approved):

TalDerei:
it'd be nice to completely get rid of this mock crate, not sure what your thinking is there about upstreaming remaining things?

turbocrime:
> it'd be nice to completely get rid of this mock crate, not sure what your thinking is there about upstreaming remaining things?

this PR totally removes `mock_ragu`

blockers that prevent alignment with `ragu/modern-deps` are
- `reddsa` dep on ff 0.13 (needs upstream work)
- `halo2_poseidon` dep on ff 0.13 (need a poseidon util in ragu's mock)

TalDerei:
> this PR totally removes `mock_ragu`

oh wait, i had local merge conflicts.

PR #127, opened 2026-05-28
fix: standardize error handling conventions

### LINKED ISSUE #82: Standardize error handling conventions
References discussion thread in https://github.com/tachyon-zcash/tachyon/pull/64#discussion_r3253255489. Today the crate mixes several incompatible error handling postures, and we should converge on a single consistent approach.

LucidSamuel:
fixes #82 
- BuildError: add `Display` + `Error` impls (was the only public error missing them)
- `ActionDigest::TryFrom`: return `ActionDigestError` instead of `&'static str`
- `verify_signatures`: return `SignatureError` with Binding/Action(idx) context instead of leaking `reddsa::Error`
- `BindingSigningKey::from_bytes`: return Result nstead of Option
- Fix stale domain byte values in `private.rs` doc comments (0x09→0x21, 0x0a→0x22)
- Add targeted tests for new error variants

### DIFF
--- a/crates/tachyon/src/keys/private.rs
+++ b/crates/tachyon/src/keys/private.rs
@@ -250,12 +250,15 @@ impl ActionSigningKey<effect::Output> {
 #[derive(Clone, Copy, Debug)]
 pub struct BindingSigningKey(#[debug(skip)] reddsa::SigningKey<reddsa::BindingAuth>);
 
-impl BindingSigningKey {
-    /// Attempt to parse a binding signing key from 32 bytes.
-    pub fn from_bytes(bytes: [u8; 32]) -> Result<Self, reddsa::Error> {
+impl TryFrom<[u8; 32]> for BindingSigningKey {
+    type Error = reddsa::Error;
+
+    fn try_from(bytes: [u8; 32]) -> Result<Self, Self::Error> {
         reddsa::SigningKey::<reddsa::BindingAuth>::try_from(bytes).map(Self)
     }
+}
 
+impl BindingSigningKey {
     /// Sign a transaction sighash with this binding key.
     pub fn sign<RNG: RngCore + CryptoRng>(
         &self,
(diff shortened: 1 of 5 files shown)


TalDerei (commented):
this LGTM, deferring to @turbocrime

TalDerei (approved):

crates/tachyon/src/keys/private.rs
```diff
@@ -250,15 +250,15 @@ impl ActionSigningKey<effect::Output> {
...
-        reddsa::SigningKey::<reddsa::BindingAuth>::try_from(bytes)
-            .ok()
-            .map(Self)
+impl TryFrom<[u8; 32]> for BindingSigningKey {
+    type Error = reddsa::Error;
+
+    fn try_from(bytes: [u8; 32]) -> Result<Self, Self::Error> {
+        reddsa::SigningKey::<reddsa::BindingAuth>::try_from(bytes).map(Self)
     }
+}
```
turbocrime, line 259: converted `from_bytes` into a `TryFrom<[u8; 32]>` impl

crates/tachyon/src/primitives/action_digest.rs
```diff
@@ -47,15 +47,6 @@ impl From<ActionDigest> for [u8; 32] {
...
 
-impl TryFrom<&[u8; 32]> for ActionDigest {
-    type Error = &'static str;
-
-    fn try_from(bytes: &[u8; 32]) -> Result<Self, Self::Error> {
-        let fp: Fp = Option::from(Fp::from_repr(*bytes)).ok_or("invalid field element")?;
-        Ok(Self(fp))
-    }
-}
-
```
turbocrime, line 58: previously, this added a case to ActionDigestError, which meant it collected digest deserialization errors with digest computation errors, which doesn't make sense for callers.

so i removed the digest conversion, since it wasn't actually used anywhere.

turbocrime (approved):
rebased and made some changes

PR #128, opened 2026-05-28
fix: prevent serialization of stripped bundles

### LINKED ISSUE #96: Prevent serialization of unassociated adjuncts
it's not valid to serialize a stripped bundle that hasn't been assigned an aggregate wtxid

possibly we should have a `Bundle<Stripped>` state before `Bundle<Adjunct>` or possibly we should require a parameter on the strip method

LucidSamuel:
fixes #96 
- adds `Bundle<Unassigned>` type-state between Stamped and Stripped.
  `strip()` now returns `Bundle<Unassigned>` which has no write() method;
  assign_wtxid() validates and transitions to Stripped.
  Keeps the `Stripped::write()` runtime guard as defense-in-depth.

### DIFF
--- a/crates/tachyon/src/bundle/mod.rs
+++ b/crates/tachyon/src/bundle/mod.rs
@@ -665,11 +665,20 @@ impl TachyonBundle {
             },
             | BundleState::Stripped => {
                 let (actions, value_balance, binding_sig) = read_bundle_body(&mut reader)?;
+                let stamp = AggregateId::read(&mut reader)?;
+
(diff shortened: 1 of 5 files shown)


turbocrime (approved):
approving with some edits

-  Renamed the `Unassigned` marker to `Stripped` and dropped the `Stamped`/`Adjunct` aliases
- Added an `AggregateId` type to constrain valid wtxids, used in `assign_wtxid`
- Updated `TachyonBundle::read` to validate deserialized bundles
- Refactored tests

PR #129, opened 2026-05-29
Deep dive into tachyon

alxiong:
run `just book serve` locally, this deep dive is meant to be read from top to bottom. I wrote in a flow that minimize forward reference (as opposes to a doc) and only introduce things incrementally. 
(only two exceptions are "Nullifier security" and "QR trick" which are self-contained and requires future concepts)

The goal for this post, is alignment on the key designs across the team. 
And I also want to eliminate "blackboxed component" as much as possible (L already implemented most of them, but there's still a few areas that is ambiguous such as how QR filters works or what is payment protocol's scope exactly or why are we Faerie Resistance). I use this writing as an exercise to convince myself that every component can work. (which Sean already knew how they work a year ago, and we've been downloading from him in piecemeal over months) 

---

update June 12th: 
~possibly helpful: non-comprehensive list of gaps with current implementation:~
~https://github.com/tachyon-zcash/tachyon/pull/129#issuecomment-4686831227~ (outdated)

update Aug 5th:
~updated divergence summary in: https://github.com/tachyon-zcash/tachyon/pull/129#issuecomment-5193440146~ (outdated)

update Aug 29th:

after rounds of consolidation, proof tree alignment, with L and Sean, the design is up-to-date reflecting our agreed shape. (not all details are reviewed yet, but general consensus on most design decisions are reached)

So, no more principal discrepancies/gaps to pay special attention to; just small details and edge case to double check.


book/src/revisit.md
```diff
@@ -0,0 +1,1289 @@
...
+
+
+Finally, the Tachyon bundle carries a spend authorization signature for every
+Action description, each signed over the `SIGHASH`, which commits to the same
+transaction-wide effecting data (across all pools) used to derive `txid`[^txid-sighash].
+Block space can additionally serve as a data-availability layer for arbitrary
+payment-protocol data used in note transmission; the shielded protocol neither
+interprets this data nor checks its correctness. As explained [later](#payment),
+the payment protocol Tachyon targets keeps KEM key material (e.g., Orchard's
+`epk`) out of band, which shrinks the in-band footprint, is quantum-safe from
```
TalDerei: the analogy here needs tightening; the KEM ciphertext + `H(ek)` is on-chain, and `ek` is OOB via PIR. in that context, there's like two analogues we can make here:
- in Orchard, `epk` is the ephemeral DH thing stored on chain, and it's an analog to the KEM ciphertext, which a recipient can read to extract a secret,
- the analogue to the KEM material kept OOB would be Orchard's `pk_d`, not `epk`

book/src/revisit.md
```diff
@@ -0,0 +1,1289 @@
...
+Re-randomization buys unlinkability by publishing a fresh-looking but valid
+key/signature each spend. With no post-quantum re-randomizable signature, we
+recover the same effect from zero knowledge. Instead of broadcasting a signature
+to be checked against $\rk$, the spender proves *knowledge* of a valid
+post-quantum signature (hash- or lattice-based) in zero knowledge. The proof
+reveals nothing about the signature, so two spends by the same key stay
+unlinkable, exactly what re-randomization provided. Since authorization is now a
+proof rather than a separate signature, it folds into the transaction's
+[PCD proof](#pq-pcd), unifying authorization and validity into one post-quantum
+artifact.
```
TalDerei: since authorization becomes a proof folded into the PCD, we should explicitly mention that `rk` will get dropped from the action description. also as a side effect, the `α = PRF(cm || θ)` binds the note to the action through `rk` (so if `rk` goes away, we should mention that binding also has to move into the proof statement).

book/src/revisit.md
```diff
@@ -0,0 +1,1289 @@
...
+$\nf_e$ for the current epoch $e$, the epoch may advance to $e+1$ before the
+transaction is picked up from the mempool. Since neither miners nor the OSS—the
+latter responsible only for syncing past epochs, and never learning future
+nullifiers, least of all at spend time—can unilaterally update the proof, the
+transaction goes stale and requires further user input to refresh. This is poor
+UX and a potential timing side-channel that leaks privacy. We therefore require
+every spend action to reveal (and prove in circuit) the nullifiers for **both the
+current and the next epoch**, leaving an ample buffer against this cross-epoch
+race. To keep spend and output actions indistinguishable, we further require each
+output action to publish a random dummy tachygram alongside its note commitment.
```
TalDerei: don't think the code's currently wiring this.

book/src/revisit.md
```diff
@@ -0,0 +1,1289 @@
...
+quantum computer learns nothing about old transactions:
+
+- the owner field $\pk = H(\ak, \nk)$ and the note commitment $\cm$ are
+  hash/symmetric (Poseidon) commitments, hiding even against a quantum computer;
+- nullifiers are PRF/hash outputs, pseudorandom against a quantum computer, so
+  [spend unlinkability](#nf-sec) survives;
+- memos travel under [ML-KEM](#address), post-quantum from day one.
+
+The only discrete-log values on chain are the randomized validating key
+$\rk = [\ask + \alpha]\,\G$ and the binding key. Re-randomization makes even
```
TalDerei: if we wanted to be complete here, we could also mention `cv`.

book/src/revisit.md
```diff
@@ -0,0 +1,1289 @@
...
+transaction goes stale and requires further user input to refresh. This is poor
+UX and a potential timing side-channel that leaks privacy. We therefore require
+every spend action to reveal (and prove in circuit) the nullifiers for **both the
+current and the next epoch**, leaving an ample buffer against this cross-epoch
+race. To keep spend and output actions indistinguishable, we further require each
+output action to publish a random dummy tachygram alongside its note commitment.
+
+All non-malleable parts, collectively the *effecting data*, hash into a stable
+identifier `txid`: a bundle commitment from each pool, their value balance
+$v^{\mathsf{bal}}$, and encrypted memo bytes.
```
TalDerei: slight contradiction with "encrypted memo bytes"; you later mention "for the Tachyon pool the memos are not effecting data and are therefore excluded from the txid inputs".

book/src/revisit.md
```diff
@@ -0,0 +1,1289 @@
...
+
+Each block contains one or more transactions. Each transaction has a `txid`,
+which commits only to its _effecting data_
+([ZIP-244](https://zips.z.cash/zip-0244)) and thus serves as the stable
+identifier for p2p gossip, and a `wtxid`
+([ZIP-239](https://zips.z.cash/zip-0239)), which additionally commits to the
+potentially malleable authorization data.
+Each transaction optionally contains a bundle of transfers from each pool:
+JoinSplit for Sprout (soon deprecated), Spend/Output for Sapling, Action for
+Orchard, and now Tachyon Action for the new Tachyon pool.
```
TalDerei, line 588: familiar with ZIP-244, but wasn't with ZIP-239. went back to investigate this split, and it seems that `wtxid` (not `txid`) is the identifier used for p2p relay of v5+ transactions per ZIP-239.

book/src/revisit.md
```diff
@@ -0,0 +1,1289 @@
...
+folds in $\nf_e$ and $\nf_{e+1}$, an output its commitment and a dummy), the
+accumulator $\tgacc$, and the anchor, the validator:
+
+1. confirms the `anchor` is a genuine node value in the consensus
+   [anchor chain](#anchor);
+2. verifies the stamp's PCD proof against $(\set{\tg_i}, \tgacc, \mathsf{anchor})$,
+   i.e. the [Action statement](#statement). The statement internally enforces
+   $\tgacc$'s consistency with the published $\set{\tg_i}$ (the
+   [batched check](#acc-correct)), the integrity of the revealed nullifiers and
+   output commitments, and the spendability of every spent note up to the anchor.
```
TalDerei, line 1127: consensus also needs some kind of like anchor recency rule I think (like anchor `epoch >= e - 1`) or there's a gap here.

currently zcash has mempool tx expiry via `nExpiryHeight`, but we're updating epochs with stamp lift operations, so we don't need to worry about that (the anchor is the expiry handle).

book/src/revisit.md
```diff
@@ -0,0 +1,1289 @@
...
+
+<P align="center">
+  <img src="./assets/tachyon_keys.svg" alt="tachyon_keys" />
+</p>
+
+The payment key $\pk = H(\ak, \nk)$ represents a note owner.
+The hash-based derivation from the $(\ak, \nk)$ tuple provides a succinct owner
+field in a note and offers
+[quantum recoverability](https://zips.z.cash/draft-ecc-quantum-recoverability) today.
+Publicizing $\ak$, a Schnorr signing key, to senders who might have future access
```
TalDerei: `ask` is the signing key, `ak` is the spend validating key (redpallas vk).

book/src/revisit.md
```diff
@@ -0,0 +1,1289 @@
...
+
+In Zcash today, a shielded payment address binds together $(\ak, \nk)$ and
+additionally includes $\ivk$ for incoming note detection. Tachyon instead
+decomposes this structure into a payment key $\pk = H(\ak, \nk)$ and a separate
+transmission key managed entirely by the payment protocol. This significantly
+simplifies the shielded protocol’s key architecture by removing functionality
+unrelated to spend authorization.
+
+> Among the main [security properties](https://zcash.github.io/orchard/design/nullifiers.html#security-properties),
+> Tachyon shielded protocol needs to uphold Ledger Indistinguishability,
```
TalDerei: we should cite the ledger indistinguishability from the zerocash paper.

TalDerei (commented):
some first pass observations, really great writeup!

book/src/revisit.md
```diff
@@ -0,0 +1,1289 @@
...
+otherwise. The $k$ filters together sort the field into $2^k$ disjoint buckets of
+roughly equal size.
+
+<a id="batch-qr">**Batched QR Test.**</a> Given the vanishing polynomial over
+some elements $f(X)=\prod_i{(X - x_i)}$, we can batch-test that all elements are
+QR, namely $\forall x_i \in \QR$, as follows:
+
+- Prover interpolates all QR pairs $(x_i, y_i)$ into a polynomial $g(X)$ where
+$g(x_i) = y_i$ and $x_i = y_i^2$
+- Prover computes $h(X)=\frac{g(X) - X}{f(X)}$ and sends commitments to $g(X)$
```
LucidSamuel: I think the divisibility relation might be incorrect as written.
given `g(x_i) = y_i `and `y_i² = x_i`, evaluating the stated numerator gives `g(x_i) - x_i = y_i - y_i²`, which is not generally zero. the polynomial that vanishes on every root is: `g(X)² - X`,  so the quotient and randomized check should be: `h(X) = (g(X)² - X)` / `f(X)` and `g(r)² - r = f(r)h(r)`.

think the offset and non-residue variants would similarly use `X + R` and `c(X + R)` since the later sibling-purity argument relies on this test, could this identity be corrected?

book/src/revisit.md
```diff
@@ -0,0 +1,1289 @@
...
+subset of the input domain and nowhere else. Section 3.3 of [BW13]
+explains the *prefix-fixing* family realized by the seminal 
+[[GGM84] PRF](https://crypto.stanford.edu/pbc/notes/crypto/ggm.html)
+which is sufficient for us.
+
+#### GGM-based Nullifier {#nf-ggm}
+
+Let $\bits{e} = (e_0, e_1, \ldots, e_{\ell-1}) \in \{0,1\}^\ell$ be the
+binary expansion of $e$ (so $e_0$ is the most significant bit).
+Let $G: \{0,1\}^s \rightarrow \{0,1\}^{2s}$ be a length-doubling PRG, split into
```
LucidSamuel: The current implementation is quaternary, not binary: GGM_TREE_ARITY = 4 and GGM_CHUNK_SIZE = 2 in crates/tachyon/src/keys/ggm.rs:24,32.

In the production configuration, EPOCH_MAX = 2²⁰ - 1, giving a tree depth of 10. End-to-end nullifier derivation performs one nf_master hash, ten nf_prefix hashes, and one final domain-separated nullifier hash. Thus the implemented cost is 12 Poseidon hashes, or ℓ/2 + 2 for ℓ = 20, rather than ℓ + 1.

The binary delegation examples below also do not carry over unchanged. could this section either describe the generic arity-A construction and instantiate A = 4, or explicitly mark the binary tree as a pedagogical simplification?

book/src/revisit.md
```diff
@@ -0,0 +1,1289 @@
...
+The [decoupling](#decouple) split the owner-binding payment key from the
+note-transmission key, leaving the latter for the payment protocol to define. A
+Tachyon full payment address is
+
+$$
+\addr = (\underbrace{\tag}_{\text{per-note}},\; \underbrace{H(\ek),\, \pk_d}_{\text{per-sender}})
+\qquad
+\begin{cases}
+    (\ek, \dk) \leftarrow \mathsf{ML\text{-}KEM.KeyGen}()\\
+    \pk_d = H(\ak, \nk; \rpk)
```
LucidSamuel: possible error on my end: the current spend relation only accepts the static payment `key pk = H(ak, nk)`. both `NfMasterSeed` and `SpendBind` require `note.pk` == `pak.derive_payment_key()`, and there is no rpk witness or diversified-key relation.

consequently, a note whose owner field is `pk_d = H(ak, nk; rpk)` would not currently be spendable. so is this intended as a future protocol change? If so, it would help to describe the additional relation the spend proof must enforce. otherwise maybe `pk_d `should probably remain payment-protocol metadata rather than replacing the note’s owner field.

LucidSamuel (commented):
WIP: Really solid piece, I read it top to bottom and never needed to forward-reference, which is exactly what the PR description promises. The QR-trick section (especially the sibling-purity footnote) and the nullifier security analysis are the highlights for me. I cross-checked the protocol description against the current crates/tachyon implementation; most of my comments below are about doc↔code alignment, plus a few Nits. Amazing work!

alxiong:
As I rewrite the entire Payment protocol section in [70fb597](https://github.com/tachyon-zcash/tachyon/pull/129/commits/70fb5979cf364c44cda18e72e46248683803edb2), 

The most important subsections are "Full Payment Address" and "Note discovery". The content there should at least 80% aligned with what Dev suggested in Tokyo. and I took the liberty to fill in more details such as the exact tag derivation, (and why some other variations don't work), and the wallet state needs to be stored encrypted onchain for full wallet recovery. 

^^ in both of those cases, i only elaborate details when I can't blackbox it while arguing it meets our security requirement. For places like how do we do first-contact exactly, I only provide options and grey-box it, as it doesn't affect the security reasoning.

book/src/revisit.md
alxiong: you are right that "Tachyon keeps KEM material OOB" is inaccurate. only the first contact post KEM ciphertext on chain, all subsequent ones in that channel just reuse the established shared secret. 

and yes, `pk_d` is kept OOB. I will update this claim.

alxiong: update: fixed in [36adaf9](https://github.com/tachyon-zcash/tachyon/pull/129/commits/36adaf9d961556cd14791ddb5af290fd4321a821)

book/src/revisit.md
alxiong: no, it's not. there's bunch of things current implement diverges from this writeup. (I'm not caught up with code, was planning to rely on AI to find all the gaps later)

i summarized gaps that i know: https://github.com/tachyon-zcash/tachyon/pull/129#issuecomment-4686831227

book/src/revisit.md
alxiong: fixed in [36adaf9](https://github.com/tachyon-zcash/tachyon/pull/129/commits/36adaf9d961556cd14791ddb5af290fd4321a821)

book/src/revisit.md
alxiong: i fix some inconsistency around `e-1` issue in: [36adaf9](https://github.com/tachyon-zcash/tachyon/pull/129/commits/36adaf9d961556cd14791ddb5af290fd4321a821)

OSS update until `e-1`, but user needs to lift it up to `e`. and yes, the final proof needs to show a spendability proof with anchor in e. which is the recency rule you want more explicitly.

book/src/revisit.md
alxiong: you are right, typo. fixed now.

book/src/revisit.md
alxiong: so, i deliberately avoid adding in `R` in these number theory background. because as you said it's simply a shift. What i'm presenting is the non-shifted case. 

notice that i said:
> ∀xi​∈QR,

without any subscript of filters, i didn't say: $\forall x_i \in \mathsf{QR}_{R_k}$

book/src/revisit.md
alxiong: yes, that's right. that's an optimization trick L had. Here i'm presenting the idea. 

Generally, we could even have a mixed-arity within the tree (which is what L ultimately wants). higher arity closer to root and arity-2 closer to leaves, so that we reduce the total height (thus total number of hashes in circuit), and enough range granularity for OSS delegation.

good to point this out, I should add an callout to explain this optimization idea succinctly.

book/src/revisit.md
alxiong: there are a few divergence between this writeup and impl. I didn't rely on any existing status or decision of existing code. 

I updated the payment key to be `pk_d` consistently in [36adaf9](https://github.com/tachyon-zcash/tachyon/pull/129/commits/36adaf9d961556cd14791ddb5af290fd4321a821), we should update the code in the future to match that. 

and again `pk_d` is not just a payment protocol thing. It should be baked into shielded protocol because the only the diversified pk should be used. the definition of an address is `addr = (pk_d, ek)` and there's no benefit of using `pk` as the note owner.

book/src/revisit.md
alxiong: updated in [3fe16dd](https://github.com/tachyon-zcash/tachyon/pull/129/commits/3fe16ddbcb4ec40c340903ab586d0e5262739943)

alxiong:
I'm unfamiliar with actual implementation status, but a few divergence that I'm aware of:

- code uses `pk`, but we need to use `pk_d` the diversified version
- nullifier opening is different, L's doc contains an extra trapdoor, which I'd argue serves no purpose afaict. So my nullifier opening is shorter. The GGM is also not exactly the same, like how we instantiate $G_0, G_1$ etc. And there's the mixed arity.
- stamp in code doesn't include the result `acc^tg` which saves validator from recomputing. similarly the proof tree didn't let validator batch-verify the accumulator correctness. 
- in code, not all Actions have two tachygrams, no dummy padding for Output action yet

book/src/revisit.md
LucidSamuel: coming back to this with the derivation written out, because I think the divisibility claim doesn't hold as currently stated and it's load-bearing for the whole QR-exclusion argument.

The witness relation in the batched QR test is quadratic: $x_i = y_i^2$ (each opened value $y_i$ is a square root of the root $x_i$). But the vanishing numerator is written as $g(X) - X$, which is linear in $g$ and it kinda doesn't add up.

check it at a root $x_i$ of $f$. we have $g(x_i) = y_i$ with $y_i^2 = x_i$, so:

$$g(x_i) - x_i = y_i - y_i^2.$$

that's zero only if $y_i \in {0, 1}$ so it's nonzero for a generic QR root. ergo $f(X) \nmid \big(g(X) - X\big)$, and the divisibility check doesn't certify residuosity,  a prover can take $g = X$ (so $g(X) - X \equiv 0$, trivially divisible) and pass while proving nothing about the $x_i$ being residues... the check never squares anything, so the $y^2 = x$ relation is never tested.

the numerator that matches the $y^2 = x$ relation is

$$g(X)^2 - X,$$

since $g(x_i)^2 = y_i^2 = x_i$ at every root, so $f \mid \big(g(X)^2 - X\big)$ exactly when each opened value really is a square root, the randomized check at $r \sample \F$ then becomes:

$$g(r)^2 - r \iseq f(r) \cdot h(r), \qquad h(X) = \frac{g(X)^2 - X}{f(X)}.$$

re: the non-shifted case, agreed the shift is orthogonal here. I think the thing that remains is that the relation is quadratic in $g$ either way ($g^2 = x$), so the numerator needs to be $g(X)^2 - X$ rather than $g(X) - X$ as it's what ties the check to the QR property and as currently written a prover could satisfy it without proving residuosity.

book/src/revisit.md
```diff
@@ -0,0 +1,1638 @@
...
+*Unlinkability.* The OSS hands back a proof anchored in $e-1$. Folding that anchor
+into the stamp *verbatim* would let the OSS recognize the very anchor it just
+produced and link the eventual spend back to the syncing session. The wallet must
+re-lift to a fresh anchor of its own to erase that handle.
+
+*Two tachygrams, not three.* Given that a lift is required, the wallet could still
+lift only *within* $e-1$, leaving the exclusion claim anchored in the previous
+epoch. But then the blind spot consensus must close with its live window would
+stretch back across $e-1$, and catching a double-spend there would force the
+wallet to also publish the $\nf_{e-1}$ — a *third* tachygram on top of
```
LucidSamuel, line 1085: this is clearer now, but just for clarity, I think the invariant should be stated explicitly, the two-epoch window checks duplicate *published tachygrams*, not “this" spend’s nullifier for each scanned epoch.”

so an epoch-`e` spend does not need to publish `nf_{e-1}` because any prior spend in `e-1` would already have published `nf_e` as its next-epoch nullifier, and the `{e-1, e}` window catches that collision. this depends on the final wallet-side lift into epoch `e` above being mandatory btw

LucidSamuel (commented):
GG; the incremental flow holds top to bottom. most of my notes are doc↔code alignment, not the design.

book/src/revisit.md
alxiong: You're referencing a line on a collapsed sidenote which only aims to provide more explanation of why alternative doesn't work.

I thought the main text is already clear about both the publishing rule and the 2 epoch window check. Do you find the current language still lacking? (The Consensus Window section)

Screenshot 2026-06-18 at 10 01 58 AM

or do you want something more explicit?

book/src/revisit.md
alxiong: you are right, that's a typo/error on my part. (i didn't notice it, i have it correct on my draft paper, but type wrong in latex.), thanks for catching it

corrected in [7d6dc31](https://github.com/tachyon-zcash/tachyon/pull/129/commits/7d6dc315619841143fd6a0693139c4fdf456fc1f)

book/src/revisit.md
LucidSamuel: ACK

alxiong:
Claude's summary of the design differences between current `proof-tree.md` and my `revisit.md`:

The core fork: how the note's identity is keyed across the OSS-proven spine.

- L (proof-tree.md) keys the spendable by the concrete per-epoch nullifier `nf`. `SpendableHeader = (nf, anchor)`, and the spend binds by checking `present_nf == spendable.nf` (proof-tree.md:31). Because `nf` evolves every epoch, this forces the whole rollover sub-machinery (`RolloverFuse/DelegateRolloverFuse/SpendableRollover/SpendableEpochLift`), and because nf is per-note the exclusion segments can't be shared.
- A (revisit.md #prooftree) keys by a constant note tag `κappa = H(seed, cm)`. `UnspentHeader = (κappa, pck, anchor, j)` carries no nf at all; each epoch's `nf_i` is derived inside `CrossEpochLift` from `pck`, tested, and not appearing in the public inputs. Constant `κappa` removes rollover entirely and lets the heavy per-epoch evidence be note-independent and shared.

Almost every gap below traces back to this. @turbocrime

alxiong:
This is a new summary of divergence of design, between L's and mine, after another scan by Fable toady. 

### Already Consolidated

Lots of pointers in https://github.com/tachyon-zcash/tachyon/pull/129#issuecomment-4686831227 are bridged/patched; even a core split (https://github.com/tachyon-zcash/tachyon/pull/129#issuecomment-4834902035) to our proof tree approaches are also consolidated. This is good news.

>The "core fork" from the June PR-129 comment (nf-keyed SpendableHeader + rollover machinery vs your κ-keyed spine) no longer exists. Upstream PRs #136/#138/#151 deleted the rollover steps and rebuilt the sync spine around exactly the delegation shape nf-analysis Attempt 3 implies: the wallet shares nullifier values, the OSS commits them into sentinel-terminated polynomial sequences (UnspentFuse/UnspentEpochFuse splice identities at Fiat–Shamir points), and the wallet binds them to genuine derivation at VerifyUnspent. Also converged since June: txid/auth_digest split (stamp is auth data), action digest Poseidon(cv, rk) accumulated as polynomial roots, α = H(θ ‖ cm) binding, two published nullifiers per spend, the two-epoch consensus window, per-stamp Poseidon anchor chain, and stripped bundles referencing the covering aggregate by wtxid.

The main consensus for our proof tree approach is that: we will have note-independent epoch evidence for cross-epoch anchor lifting that's shareable by all. 

### Changes to make

I concede the following changes are needed for my current writeup.

Done:
- Flip the value sign convention to Zcash's (spend `+v` / output `−v`, `v_bal = Σspend − Σout`)
- Add the `1 ≤ v ≤ MAX_MONEY` range check to the statements
-  Replace the Pedersen action accumulator `acc^act` inside `txid` with an order-committing descriptor hash
-  Add `nf != 0` / `cm != 0` / non-identity-point guards to the Spend/Output statement
- Add a note about rerandomization during hand-off of proofs across trust boundaries
- Fixed the output authorization being `rk = [alpha]G` rather than `ak + [alpha]G` (no authorization needed, this is my typo, L's code and doc are correct)

In progress:

- [x] replace GGM with (generic) poly-based nullifiers, preferably with Attempt 3 in [this analysis](https://github.com/alxiong/tachyon/pull/1) (namely single Poseidon per nf), introduce a generic interface
- [x] update proof tree to reflect the poly-based nullfiiers, in particular, the OSS needs to "accumulate" claimed `nf` that he proved exclusion for, then later "unrolled" by the readers who prove their derivation in batch. (this is already in L's impl, but my writeup outdated in this regard)
- [x] update any security argument related to or implicated by `nf` design. 
- Full mempool aggregation protocol. (i need to re-review #148 and update my writeup/semi-spec)

### Remaining Differences for Discussions

1. ~~**diversified payment key**. I argued we need diversified payment key: `pk_d​=Com(ak, nk; rpk)`. I explained the overall diversified address (with transmission key) in "Payment protocol" section; I believe we cannot "push-off" the decision/responsibility of enforcing payment key diversification to payment protocol; it has to be enforced in our action statement.~~
2. **In-band DA/Memo**. As I worked on the full process of note discovery, and its requirement to be able to recover from mnemonic, we have to record first-contact's ciphertext on-chain. Thus we do need optional opaque DA blobs in Tachyon tx, and `da_digest` field, along side `auth_digest` and `wtxid`. 
3. **Proven `acc^tg` to avoid recompute**. In a private convo with Sean months ago, he suggested including the tachygram accumulator `acc^tg` in each stamp, and prove its integrity via cheap batched verification (that only involves field ops) and save validator from recomputing it. 
   - meanwhile I'd argue recomputing `acc^act` is fine because it was pure hashes. no group ops.
4.  **Anchor chain: tick frequency and epoch binding.**
    - In L's design, even an empty block tick the anchor chain; in mine, I only tick per stamp, so block without any tachyon bundle won't tick anchor chain forward. IMO, both are fine, with slight bias to mine for its simplicity and don't have to detect "end-of-block" condition to know to tick
    - my anchor state `st_new = H(st_old ‖ e ‖ tgacc)` binds to epoch on *every* tick/stamp, I added this because I realize i need this for my proof tree. Whereas L only bind the epoch number at the epoch boundary. I find mine easier to reason about in cases where we need to assure "this mid-epoch anchor belongs to epoch `e`"; which is something we need when doing in-epoch lifting.
5. **Tachygram padding**. Mine: all actions, spend or output, produces two tachygrams, padded with dummy if needed. I think L has some pending PR to fix this? 
6. **Zero-value Note allowed**. currently only `[0, MAX_MONEY]` allowed (https://github.com/tachyon-zcash/tachyon/blob/a0abbdaae72648aa260274c5ad5acc1a1c6218e8/crates/tachyon/src/note.rs#L89), but we should follow Orchard's convention. perfectly fine to have zero-value note imo. 

---

Update: Aug 6th. 

Some feedback from @ebfull (cc @turbocrime):

- remove "diversification" altogether, just call it **fresh address** (per sender), since we lose persistent `ivk` and technically incorrect to call it diversified address/payment_keys. 
- in-band DA confirmed,  `txid` needs to bind `da_digest` DA bytes, to avoid any malleability.
- tachygram accumulator wanted, and confirm that proving it to save validator's work. Sean's mental model is client/user provide "advice" so that validator can quickly check, which is more efficient than recomputing the commitment. But in practice, this is effectively the same as what I propose: merging into the PCD proof, and we prove it with cheap poly evals in circuit and a commitment opening check via the poly query oracle.
- anchor chain: Sean's original push for "tick for empty block" is to avoid complication when the last block of an epoch not carrying a tachyon bundle, thus the cross-epoch tick in the epoch chain is missing. BUT, if we bind each anchor update with the current epoch `e`, then we completely avoid this concern. I also convey to Sean that in my Proof tree design, in-epoch lifting is a different step than cross-epoch lifting. The former restrict the start and end anchor to be in the same epoch, and only the latter allows and checks e+1. In short, I still hold that my design is fine.
- spend/output indistinguishability is desirable, thus padding is needed. the number of signature will reveal `nActions`, and with `nTachygrams`, the arity of spend/output is leaked, although one don't know which is cm/nf. Sean points out we should use a separate domain-separated hash of a garbage/random preimage for dummy note commitment instead of zero-value note, which I agree.

book/src/revisit.md
```diff
@@ -0,0 +1,2314 @@
...
+tachygrams, the accumulator of that union, the common anchor, and an aggregated
+PCD proof. The stamps of all constituent transactions are replaced by a
+reference to the aggregated transaction's `wtxid`.
+
+> Note: an aggregated Tachyon bundle shares exactly the same format as a normal
+> standalone bundle (a.k.a. a _Tachyon autonome_), and may even carry additional
+> Action descriptions of its own. A purely aggregating bundle, by contrast,
+> carries an empty Action list (hence no authorization signatures), a zero value
+> balance, a trivial binding signature, and a stamp holding the aggregated proof
+> and its proof data.
```
TalDerei: what's the context for a "purely aggregating bundle" that doesn't carry any transfers of its own? seems like a degenerate case of an aggregate.

turbocrime (approved):

PR #132, opened 2026-06-03
zeroize secret types on drop

### LINKED ISSUE #131: Zeroize secret key and trapdoor types on drop
Several secret-bearing types do not zeroize their contents when dropped. After #68 added opaque `Debug` impls, zeroization is the next layer of secret hygiene.

## Affected types

**Keys** (`keys/private.rs`):
- `SpendingKey([u8; 32])`
- `SpendAuthorizingKey(reddsa::SigningKey<SpendAuth>)`
- `ActionSigningKey<E>(reddsa::SigningKey<SpendAuth>)`
- `BindingSigningKey(reddsa::SigningKey<Binding>)`

**Trapdoors** (`note.rs`, `value.rs`):
- `note::NullifierTrapdoor(Fp)`
- `note::CommitmentTrapdoor(Fp)`
- `value::CommitmentTrapdoor(Fq)`

All are currently `Clone + Copy`. Since `Drop` types cannot be `Copy`, implementing `ZeroizeOnDrop` requires removing `Copy` from these types and fixing call sites.

## Proposed direction

1. Add `zeroize` as an explicit dependency.
2. Implement `Zeroize` and `ZeroizeOnDrop` for the wrapper types listed above.
3. Remove `Copy` from secret-bearing types; update call sites to use references or explicit `.clone()`.
4. Choose a defensible zeroization strategy for inner types that don't expose `Zeroize` directly (`pasta_curves::Fp`/`Fq`, `reddsa::SigningKey`) likely zeroize the byte representation or use `unsafe` zeroize on the raw memory with a safety comment.

Extends #60 / PR #68.

LucidSamuel:
Fixes #131 

adds zeroize-on-drop handling for tachyon secret types.

changes:
 - Adds `zeroize` as an explicit dependency. It is already present transitively via `reddsa`.
 - Implements `Zeroize` + `Drop`/`ZeroizeOnDrop` for:
   - `SpendingKey`
   - `SpendAuthorizingKey`
   - `ActionSigningKey<E>`
   - `BindingSigningKey`
   - `note::NullifierTrapdoor`
   - `note::CommitmentTrapdoor`
   - `value::CommitmentTrapdoor`
 - removes `Copy` from trapdoor types so secret scalars are not silently  duplicated.
 - updates affected call sites to clone explicitly where ownership is required.
 - changes value-commitment helpers to borrow trapdoors where possible, reducing unnecessary clones.
  - derives binding signing keys by summing borrowed trapdoors directly, avoiding intermediate copied scalar buffers.


crates/tachyon/src/keys/private.rs
```diff
@@ -276,29 +277,150 @@ impl BindingSigningKey {
...
+    unsafe_code,
+    reason = "volatile zeroization of non-Zeroize inner types"
+)]
+unsafe fn zeroize_flat<T>(val: &mut T) {
+    let ptr: *mut u8 = ptr::from_mut(val).cast();
+    let len = size_of::<T>();
+    // Safety: `T` is plain data with no internal pointers; byte-level
+    // zeroization is sound and the value is being dropped immediately after.
+    unsafe { slice::from_raw_parts_mut(ptr, len) }.zeroize();
+}
```
turbocrime: i'd prefer a utility that is type-scoped to the values that actually require this approach

crates/tachyon/src/value.rs
```diff
@@ -54,14 +55,19 @@ impl CommitmentTrapdoor {
         Self(Fq::random(rng))
     }
 
+    /// Borrow the inner scalar for use in arithmetic (e.g. BSK summation).
+    pub(crate) const fn scalar(&self) -> Fq {
+        self.0
+    }
```
turbocrime: prefer to name methods like this `inner`

turbocrime (changes requested):
i don't like the addition of `unsafe` but it seems unavoidable given upstream.

we should drop the free function `zeroize_flat` and replace it with specific implementations on each type that requires the unsafe zeroization.

turbocrime:
it seems like upstream `reddsa` *intended* to provide zeroize features, but the implementation was broken at some point. i wonder if it's worth making an upstream PR.

LucidSamuel:
> it seems like upstream `reddsa` _intended_ to provide zeroize features, but the implementation was broken at some point. i wonder if it's worth making an upstream PR.

 I looked into it and it seems the zeroize feature gate exists in their Cargo.toml but the actual Zeroize impl on SigningKey was either never wired up or got lost in a refactor. Probably a quick fix upstream.

TalDerei (commented):
halo2 / orchard don't pull in zerosize, zebra does transitively, and librustzcash does. seems like proving/protocol crates don't zeroize, only the wallet layer. in that context, maybe it'd be prudent to have a separate crate that defines the wallet key material (separately from like the other protocol primitives), and then have that layer do the zeroize guarding?

I'd upstream fixes to upstream reddsa / pasta curves, then we can extend zeroize to other keys / trapdoors this PR doesn't address, and have some explicit unit tests.

LucidSamuel:
re-verified and makes sense on all four points. 
one sequencing question before I restructure: I’m inclined to keep this PR focused on introducing the wallet-key crate, moving the currently covered secret types into the appropriate layer, and adding explicit zeroization tests. I’d send the reddsa and pasta_curves fixes upstream separately, then extend coverage to the remaining keys and trapdoors once those implementations are available.

because that would keep the structural change reviewable without blocking it on upstream merges. does that scope work, or would you prefer this PR to wait and include complete key and trapdoor coverage?
@TalDerei

turbocrime:
experimented with this today https://github.com/turbocrime/tachyon/tree/security/zeroize-secret-types-alternate but i'm still not happy with results

i think this isn't worthwhile until we have a reliable way to handle zeroization of `pasta_curves` types

turbocrime:
we're not going to attempt this, unless upstream provides zeroize support.

PR #126, opened 2026-05-28
Replace GGM-tree nullifier derivation with a polynomial accumulator + lift

turbocrime:
Prototype for #97.

Replaces GGM-tree nullifier derivation with a polynomial-vanishing construction. Each note carries a recipient-secret polynomial $R$, committed as

$$\psi = \sum_i R[i] \cdot G_i$$

and folded into the note commitment $\mathsf{cm}$. The epoch $e$ nullifier is a coefficient of $R$ shifted by $\mathsf{cm}$,

$$\mathsf{nf}_e = R[e - e_\mathsf{cm}] + \mathsf{cm}$$

## Spendable shape

`SpendableHeader = (future, anchor)`, where $\mathsf{future}$ commits to the shifted unconsumed tail from epoch $e$, trapdoored with $\mathsf{cm}$:

$$\mathsf{future} = \psi_e + \mathsf{cm} \cdot \left(\sum_i G_i + H\right)$$

where the epoch $e$ tail commits to

$$\psi_e = \sum_i R[(e - e_\mathsf{cm}) + i] \cdot G_i$$

Pedersen-independence of $H$ from the $G_i$ pins both the polynomial and $\mathsf{cm}$ in one point.

`SpendableLift` (wallet-only) consumes one composed `Unspent` and confirms its polynomial is a prefix of the spendable's tail via `enforce_poly_shifted_sum`, re-trapdooring the shrunken commit with the same $\mathsf{cm}$ (the consumed length stays internal to the relation). `SpendBind` witnesses $R$ as `consumed_head` ‖ `remaining_tail`, reassembles it by the same relation, preimage-confirms $\mathsf{cm}$ (which binds $\mathsf{pk}$, $\mathsf{value}$, and $\psi$), matches the recomputed $\mathsf{future}$, and reads $\mathsf{nf}_e$ and $\mathsf{nf}_{e+1}$ off the tail's first two coefficients.

## Step count: 24 → 14

- **Deleted**: 7 GGM steps + `RolloverFuse` + `DelegateRolloverFuse` + `SpendableRollover` + `SpendableEpochLift` (and headers).
- **Added**: `UnspentEpochFuse` (cross-epoch `Unspent` composition).
- **Primitives**: raw $R$ (`PronullifierSeqPoly`) vs shifted nullifiers (`NullifierSeqPoly`); `Note.psi` is a Pedersen commitment (Vesta point) digested into $\mathsf{cm}$.

PR #133, opened 2026-06-04
fix: seperate poseidon domain for nullifier m-key derivation

LucidSamuel:
fixes nullifier domain separation by giving GGM root derivation its own Poseidon domain tag.

previously, `nf_master` and `nf_prefix` both used `Tachyon-NfPrefix` even though they hash different protocol objects at the same Poseidon arity:

- `nf_master(psi, nk)` derives the per-note GGM root / master key.
- `nf_prefix(prefix_prev, step)` derives a child node during the GGM walk.

that meant root derivation and tree-step derivation were distinguished only by input meaning, not by the hash domain.

### DIFF
--- a/book/src/bundle.md
+++ b/book/src/bundle.md
@@ -50,16 +50,30 @@ The proof establishes:
 - tachygrams are correctly bound to action keys
 - action balance effect matches pool balance effect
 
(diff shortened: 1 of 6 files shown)


LucidSamuel:
closing this cos the GGM nullifier derivation this fixes is being replaced by the polynomial accumulator in #126. the domain separation issue won't exist in the new design.

PR #134, opened 2026-06-04
fix: implement safe value arithmetic types

### LINKED ISSUE #17: Safe value arithmetic types
Value balance uses bare `i64` and note values use a bounded `u64` wrapper. There are no typed wrappers providing assurances in intermediate calculations when summing across actions.

## Needed

- `ValueSum` newtype (i128 or similar, for summing across actions)
- Checked arithmetic operators (Add, Sub, Sum) on value types
- `BalanceError` for overflow
- `Sign` enum or `magnitude_sign()` for decomposition

## Current state

- `note::Value(u64)` in `crates/tachyon/src/note.rs:135` -- wraps u64 with bounds check (`assert!(value <= NOTE_VALUE_MAX)`) via `From<u64>`
- `note::Value` has `Into<i64>` and `Into<u64>` but no arithmetic operators
- `bundle.value_balance` is bare `i64` -- `crates/tachyon/src/bundle.rs:41`
- `BundlePlan.value_balance` is also bare `i64` -- `crates/tachyon/src/bundle.rs:232`
- `value.rs` has `CommitmentTrapdoor` and `Commitment` with `Add`, `Sub`, `Sum` on `Commitment` (curve point arithmetic) -- but these are for value commitments, not plaintext value arithmetic
- `Commitment::balance(value: i64)` takes bare `i64` -- `crates/tachyon/src/value.rs:151`
- No `ValueSum`, no `BalanceError`, no checked arithmetic for plaintext values

## Orchard reference

`value.rs`: `NoteValue(u64)`, `ValueSum(i128)`, `Sign`, `BalanceError`, `MAX_NOTE_VALUE`, `VALUE_SUM_RANGE`. Full checked arithmetic with `impl Add<NoteValue> for ValueSum`, `impl Sum<NoteValue> for ValueSum`.

LucidSamuel:
fixes #17 
- replace From<u64> for Value with TryFrom<u64> + Value::new() checked constructor
- add ValueError enum (Zero, Overflow) for out-of-range values
- add ValueSum(i128) with checked Add<Value>/Sub<Value> returning Option
- add Sign enum and ValueSum::magnitude_sign() for decomposition
- add BalanceError and ValueSum::to_i64() for wire-format conversion
- update Plan::commitment() to return Result<[u8; 64], CommitError>
- add CommitError unifying ActionDigestError and BalanceError
- add SignError::BalanceOverflow variant for propagation
- migrate all call sites from Value::from() to Value::try_from()

### DIFF
--- a/crates/tachyon/src/note.rs
+++ b/crates/tachyon/src/note.rs
@@ -33,7 +33,7 @@
 //! enters the polynomial accumulator. The concrete commitment scheme
 //! (e.g. Sinsemilla, Poseidon) depends on what is efficient inside
 //! Ragu circuits and is TBD.
-use core::{fmt, iter, ops};
+use core::{fmt, ops};
 
 use ff::Field as _;
 use pasta_curves::Fp;
@@ -168,12 +168,6 @@ impl Value {
         }
         Ok(Self(value))
     }
-
-    /// Returns the inner `u64`.
-    #[must_use]
-    pub const fn inner(self) -> u64 {
-        self.0
-    }
 }
 
 impl TryFrom<u64> for Value {
@@ -190,25 +184,24 @@ impl From<Value> for u64 {
     }
 }
 
+#[expect(
+    clippy::expect_used,
+    reason = "Value construction enforces NOTE_VALUE_MAX < i64::MAX"
+)]
 impl From<Value> for i64 {
-    #[expect(
-        clippy::as_conversions,
-        clippy::cast_possible_wrap,
-        reason = "NOTE_VALUE_MAX (2.1e15) < i64::MAX (9.2e18), so wrapping cannot occur"
-    )]
     fn from(value: Value) -> Self {
-        value.0 as Self
+        Self::try_from(value.0).expect("Value invariant guarantees it fits in i64")
     }
 }
 
 /// Signed sum of note values across actions.
 ///
 /// Spends contribute positive values, outputs contribute negative.
-/// The valid range is `-(max * action_count)..=(max * action_count)`,
-/// but `i128` provides ample headroom for any realistic bundle.
+/// Uses `i128` internally to provide additional headroom while accumulating
+/// values. Checked arithmetic still reports overflow, and conversion to the
+/// wire-format `i64` fails if the final balance is out of range.
 ///
-/// Use [`ValueSum::to_i64`] to convert to the wire-format `i64`
-/// for `value_balance`, which checks the result fits.
+/// Use `i64::try_from(sum)` to narrow to the wire-format `i64`.
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
 pub struct ValueSum(i128);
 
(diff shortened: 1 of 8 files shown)


crates/tachyon/src/note.rs
```diff
@@ -159,6 +190,78 @@ impl From<Value> for u64 {
...
+impl From<Value> for i64 {
+    #[expect(
+        clippy::as_conversions,
+        clippy::cast_possible_wrap,
+        reason = "NOTE_VALUE_MAX (2.1e15) < i64::MAX (9.2e18), so wrapping cannot occur"
+    )]
+    fn from(value: Value) -> Self {
+        value.0 as Self
+    }
+}
```
turbocrime, line 151: can we avoid this cast?

crates/tachyon/src/note.rs
```diff
@@ -159,6 +190,78 @@ impl From<Value> for u64 {
...
+            .map(Self)
+            .ok_or(BalanceError)
+    }
+}
+
+impl iter::Sum<Value> for Result<ValueSum, BalanceError> {
+    fn sum<I: Iterator<Item = Value>>(mut iter: I) -> Self {
+        iter.try_fold(ValueSum::ZERO, |sum, val| sum + val)
+    }
+}
```
turbocrime: this appears to be unused. should `Plan::value_balance` use it?

crates/tachyon/src/note.rs
```diff
@@ -159,6 +190,78 @@ impl From<Value> for u64 {
...
+/// Signed sum of note values across actions.
+///
+/// Spends contribute positive values, outputs contribute negative.
+/// The valid range is `-(max * action_count)..=(max * action_count)`,
+/// but `i128` provides ample headroom for any realistic bundle.
+///
+/// Use [`ValueSum::to_i64`] to convert to the wire-format `i64`
+/// for `value_balance`, which checks the result fits.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct ValueSum(i128);
```
turbocrime: Should `ValueSum` be `i128` if only `i64` are valid?

crates/tachyon/src/note.rs
```diff
@@ -159,6 +190,78 @@ impl From<Value> for u64 {
...
+
+impl ValueSum {
+    /// The zero sum (identity for addition).
+    pub const ZERO: Self = Self(0);
+
+    /// Convert to `i64` for the wire format, or error if the sum
+    /// does not fit.
+    pub fn to_i64(self) -> Result<i64, BalanceError> {
+        i64::try_from(self.0).map_err(|_err| BalanceError)
+    }
```
turbocrime: why `to_i64` instead of `impl TryFrom<ValueSum> for i64`

crates/tachyon/src/note.rs
```diff
@@ -122,34 +122,65 @@ pub struct Note {
...
+            return Err(ValueError::Overflow);
+        }
+        Ok(Self(value))
+    }
+
+    /// Returns the inner `u64`.
+    #[must_use]
+    pub const fn inner(self) -> u64 {
+        self.0
+    }
```
turbocrime: do we need this when `impl From<Value> for u64` exists, just a few lines below?

turbocrime (changes requested):

crates/tachyon/src/note.rs
LucidSamuel: kept i128 for additional intermediate accumulation headroom. Individual values fit in `i64`, but a bundle’s intermediate sum may not. 
also, Arithmetic remains checked, and the final wire-format balance is validated through `TryFrom<ValueSum>` for `i64`.

turbocrime (approved):

TalDerei (approved):

PR #146, opened 2026-06-24
fix(bundle): restore action index in RkMismatch

LucidSamuel:
follow ups to #142 
- change RkMismatch to { index, rk }, preserving the offending key while restoring its global action position. Includes regression coverage for output indexing.

- Replace derive_more’s full feature with the six derives actually used.

### DIFF
--- a/crates/tachyon/src/bundle/mod.rs
+++ b/crates/tachyon/src/bundle/mod.rs
@@ -227,8 +227,16 @@ pub enum SignatureError {
 #[non_exhaustive]
 pub enum SignError {
     /// The derived rk does not match the stored rk.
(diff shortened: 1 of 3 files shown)


crates/tachyon/src/bundle/mod.rs
```diff
@@ -387,12 +398,15 @@ impl Plan {
             });
         }
 
-        for plan in &self.outputs {
+        for (idx, plan) in self.outputs.iter().enumerate() {
```
turbocrime, line 405: the other loop uses `index` instead of `idx`, would be nice if it was consistent.

turbocrime (approved):
defer derive_more narrowing until later.

Issue #106, opened 2026-05-24
ZIP: Tachyon Aggregator Protocol


TalDerei:
**Book entry:** [`book/src/zips/tachyon-aggregator.md`](https://tachyon.z.cash/_book/zips/tachyon-aggregator.html)

## Summary

Defines aggregator nodes (which sit between mempool and miners) and how they handle `Autonomes / Aggregates / Adjuncts` stamps. Also defines miners' roles and the block construction protocol.

## Classification

| Field    | Value      |
| -------- | ---------- |
| Role     | Additive   |
| Category | Network    |

## Milestones

- [ ] Draft outline in book with design considerations
- [ ] Submitted as PR to `zcash/zips`
- [ ] Status → Draft
- [ ] Status → Proposed
- [ ] Implemented in `librustzcash` / `zebra`
- [ ] Status → Implemented
- [ ] Status → Final (post-activation)

### Zebra Implementation

- [ ] Implementation drafted

Issue #104, opened 2026-05-24
ZIP: Tachyon Bundle / Aggregate Transaction Format


TalDerei:
**Book entry:** [`book/src/zips/tachyon-bundle.md`](https://tachyon.z.cash/_book/zips/tachyon-bundle.html)

## Summary

Defines the bundle wire format, stamped/stripped variants, action structure (cv, rk, sig), value balance, binding signature, internal layout, and consensus rules for the Tachyon bundle that goes on-chain. In many ways, it's the Tachyon equivalent of [ZIP 225](https://zips.z.cash/zip-0225) (V5 tx format) / [ZIP 230](https://zips.z.cash/zip-0230) (withdrawn V6). This depends on https://github.com/tachyon-zcash/tachyon/issues/110. 

## Classification

| Field    | Value             |
| -------- | ----------------- |
| Role     | Additive / Update |
| Category | Consensus         |

## Milestones

- [ ] Draft outline in book with design considerations
- [ ] Submitted as PR to `zcash/zips`
- [ ] Status → Draft
- [ ] Status → Proposed
- [ ] Implemented in `librustzcash` / `zebra`
- [ ] Status → Implemented
- [ ] Status → Final (post-activation)

### Zebra Implementation

- [ ] Implementation drafted

PR #148, opened 2026-06-26
Aggregation protocol ZIP

turbocrime:
Adds the ZIP draft page for the Tachyon Aggregator Protocol #106

The draft specifies:

- An 8-step aggregation lifecycle from autonome publication through block validation,
  with block validation ordered fail-fast (tachygram uniqueness, adjunct association,
  action set commitment, then proof verification).
- Tachygram distinctness at three levels: within each stamp, across a block (as the
  multiset union of stamp `vTachygrams`), and as a mempool acceptance rule.
- Transaction identifier and relay semantics: stable `txid` across authorization forms,
  mandatory `MSG_WTX`, proof verification before relay using the stamp's own header data,
  and stripped bundles forbidden from the mempool.
- Covered-transaction identification: tachygram matching narrows candidates, and the
  `cActionsTachyon` commitment confirms completeness without proof verification.

Consensus rules not specific to aggregation (signatures, value balance, anchor
membership, epoch-window tachygram reuse) are cited to their owning ZIPs, not
re-specified. Remaining open questions (category, digest-amendment home, deployment,
duplicate-rule ownership) are called out in Design Considerations.

Please provide suggestions and amendments.

### DIFF
--- a/book/src/zips/tachyon-aggregator.md
+++ b/book/src/zips/tachyon-aggregator.md
@@ -226,8 +226,12 @@ the bundle's stamp and set the `tachyonAggregateId` field to the covering
 aggregate's `wtxid`.
 
 A stripped innocent (a former innocent aggregate) contributes no actions, but
-the stripped form still names a covering transaction: its `tachyonAggregateId`
-MUST refer to the aggregate that ultimately absorbed its stamp.
+the stripped form still names a covering transaction. Its `tachyonAggregateId`
+MUST identify a stamped transaction in the same block, and SHOULD refer to the
+aggregate that ultimately absorbed its stamp. The latter is unverifiable: a
+bundle with no actions contributes no action digests to any reconstruction, so
+a validator cannot distinguish the absorbing aggregate from any other stamped
+transaction (Step 8).
 
 ### Step 8: Block validation
 
@@ -255,7 +259,10 @@ governed by the duplicate-tachygram consensus rule of the
 2. **Adjunct association.** The `tachyonAggregateId` of every stripped bundle
 MUST identify a stamped transaction in the same block. If the referenced
 transaction is absent from the block, or bears no stamp, the validator MUST
-reject the block.
+reject the block. A stripped bundle with no actions satisfies this check
+against any stamped transaction in the block: it contributes no action digests
+to the reconstruction in item 3, so consensus attaches no further meaning to
+its reference.
 3. **Action set commitment per stamp.** The validator MUST confirm each stamp's
 `cActionsTachyon` by reconstruction: collect the action digests of the stamped
 bundle's own actions together with those of every stripped bundle that names it
@@ -409,7 +416,10 @@ mempool can identify covered autonomes by the `cActionsTachyon` check (see
 [Covered-transaction identification](#covered-transaction-identification)). This
 is inherent to the scheme: the aggregate must carry enough information for validators
 to reconstruct its header. The tachygram-set and action-digest-set commitments do not
-reveal the private contents of any covered note.
+reveal the private contents of any covered note. The `tachyonAggregateId` of a stripped
+innocent is not consensus-validated beyond naming a stamped transaction (Step 8, item 2),
+so an observer reconstructing aggregation relationships cannot rely on an actionless
+bundle's reference.
 
 **Circuit/consensus boundary.** Several security properties are enforced by consensus
 rather than fully proven inside the stamp. Double-spend prevention rests on the block-scoped
--- a/book/src/zips/tachyon-aggregator.md
+++ b/book/src/zips/tachyon-aggregator.md
@@ -43,6 +43,12 @@ Open questions:
 - **Coverage-query relay message.** A reviewer proposes a `getxref`-style relay message
   that returns the bundles whose stamp references a given `wtxid`, with nodes maintaining
   a coverage graph to answer it. Whether to add such a message is undecided.
+- **Merge-input acquisition.** An aggregator selecting an existing aggregate must recover
+  the contributing actions, but contributors cannot be requested by `wtxid` because the
+  aggregate does not carry their identities. The draft's present position is that
+  aggregators index mempool data and do not attempt merges they cannot satisfy from data
+  they hold. Alternatives: a relay message requesting transactions by tachygram, or
+  aggregates carrying a list of contributing transactions. Undecided.
 - **Stripped zero-action coverage.** A stripped bundle with no actions names a covering
   aggregate that consensus does not confirm; candidate confirmation rules are an open
   question of the
@@ -184,8 +190,9 @@ so this guidance needs no enforcement (see the [Rationale](#rationale)).
 ### Step 3: Witness and PCD preparation
 
 The aggregator reconstructs each selected transaction's stamp PCD from the
-proof and data carried on that transaction itself, obtained in a single
-`MSG_WTX` getdata. Reconstruction requires no multi-stage witness fetching.
+proof and data carried on that transaction itself, which the aggregator
+already holds from mempool observation (Step 2). Reconstruction requires no
+multi-stage witness fetching.
 
 Merging is defined only over stamps bearing identical anchors. If the selected
 transactions bear unequal anchors, the aggregator first aligns them by lifting
@@ -196,16 +203,17 @@ If both selected transactions are autonomes, all necessary witness data is
 directly available on the transactions themselves.
 
 A selected transaction that is already an aggregate additionally requires the
-action digests of every transaction contributing to it. Recovering those
-contributors is
+action digests of every transaction contributing to it. The contributors
+cannot be requested from the network by `wtxid`, because the aggregate does
+not carry their identities. Recovering them from transactions the aggregator
+holds is
 [covered-transaction identification](#covered-transaction-identification): a correct and
 complete collection of action digests reproduces the action set commitment on the
 selected stamp.
 
-Aggregators typically maintain recent consensus data, cached anchor lift
-proofs, and an index of recent mempool transactions to support alignment and
-witness preparation. These are implementation concerns; this ZIP does not
-constrain them.
+Aggregators therefore maintain an index of recent mempool transactions, along
+with recent consensus data and cached anchor lift proofs, and do not attempt
+merges whose contributors they cannot recover from data they hold.
 
 ### Step 4: Aggregate construction
 
--- a/book/src/zips/tachyon-aggregator.md
+++ b/book/src/zips/tachyon-aggregator.md
@@ -58,64 +58,64 @@ The term "network upgrade" is to be interpreted as described in
 [ZIP 200](https://zips.z.cash/zip-0200). The terms "Testnet" and "Mainnet" are to be
 interpreted as described in section 3.12 of the Zcash Protocol Specification.
 
-- **Tachygram.** A `byte[32]` field element ($\mathbb{F}_p$) representing either a note
-  nullifier or a note commitment. Consensus treats nullifiers and commitments
-  identically.
-- **Stamp.** A bundle trailer which may replaced with a reference to another
-  transaction. Contains a Ragu proof and some associated data necessary for
-  verification, including tachygrams.The proof establishes that the actions follow
-  the correct rules.
-- **Autonome.** A stand-alone transaction with a stamped bundle, containing a
-  proof covering only its own actions. The standard form of a user-originated
-  Tachyon transaction. An autonome may appear in a block or in the mempool.
-- **Aggregate.** A transaction with a stamped bundle, containing a merged stamp
-  that covers other transactions. More specifically, an *innocent aggregate*
-  contains no Tachyon actions of its own; and a *based aggregate* contains Tachyon
-  actions. An aggregate may appear in a block or in the mempool.
-- **Adjunct.** A transaction with a stripped bundle. Its proof has been removed
+The terms "tachygram" and "stamp" are defined by the
+[Tachyon Shielded Protocol](tachyon-shielded-protocol.md) and
+[Tachyon Bundle / Aggregate Transaction Format](tachyon-bundle.md) ZIPs respectively, and
+are summarized here non-normatively. The remaining terms are defined by this ZIP.
+
+- **Tachygram.** The `byte[32]` encoding of a field element ($\mathbb{F}_p$)
+  representing either a note nullifier or a note commitment. Consensus treats
+  nullifiers and commitments identically.
+- **Stamp.** A bundle trailer carrying a Ragu proof and the data necessary to
+  verify it, including tachygrams. The proof attests that every covered action
+  satisfies the Tachyon action rules. A bundle's stamp may be replaced by a
+  reference to a covering transaction (see Adjunct).
+- **Autonome.** A stand-alone transaction with a stamped bundle, whose proof
+  covers only its own actions. The standard form of a user-originated Tachyon
+  transaction. An autonome may appear in a block or in the mempool.
+- **Aggregate.** A transaction with a stamped bundle, whose merged stamp covers
+  other transactions. An *innocent aggregate* contains no Tachyon actions of its
+  own; a *based aggregate* contains Tachyon actions. An aggregate may appear in
+  a block or in the mempool.
+- **Adjunct.** A transaction with a stripped bundle: its stamp has been removed
   and replaced by a `wtxid` reference to a covering aggregate. Adjuncts retain
-  action data, action signatures, the binding signature, and `value_balance`. An
-  adjunct should only appear in a block.
+  action data, action signatures, the binding signature, and `value_balance`.
+  An adjunct is valid only within a block.
 
 ## Abstract
 
-Tachyon shielded transactions use a recursive proof system. To reduce proof
-verification costs, recursion is used to combine many per-transaction proofs
-into a single proof covering all contributing transactions.
+Tachyon shielded transactions use a recursive proof system. Recursion allows
+many per-transaction proofs to be combined into a single proof covering all
+contributing transactions, reducing proof verification costs. This recursion
+admits a new participant role, the aggregator, without creating a new trust
+assumption.
 
-This recursion provides an opportunity to introduce a new participant role
-without creating a new trust assumption.
-
-This ZIP specifies: the aggregator protocol, a block-layout discipline under
-which miners strip redundant proofs, the semantics about effecting data and
-authorizing data that make stripping safe, and P2P rules that extend ZIP 239 to
-Tachyon's authorization-form malleability.
-
-Aggregation involves an 8-step lifecycle from transaction authorization, through
-the mempool, to block layout, and final validation. All effecting data (actions,
-value balances, action signatures, and binding signatures) remains present and
-valid when a proof is stripped.
+This ZIP specifies the aggregator protocol: an 8-step lifecycle from
+transaction authorization, through the mempool, to block layout and final
+validation. It comprises a block-layout discipline under which miners strip
+redundant proofs, the effecting-data and authorizing-data semantics that make
+stripping safe, and P2P rules extending ZIP 239 to Tachyon's
+authorization-form malleability. All effecting data (actions, value balances,
+action signatures, and binding signatures) remains present and valid when a
+proof is stripped.
 
 ## Motivation
 
-Every proof must be verified to reach consensus. Without aggregation, every
-Tachyon bundle would carry a stamp, and consensus costs would be dominated by
-stamp data and verification. By specifying an aggregation protocol, consensus
-avoids this complexity.
+Consensus requires every proof in a block to be verified. Without aggregation,
+every Tachyon bundle would carry a stamp, and consensus costs would be
+dominated by stamp data and stamp verification. Aggregation bounds that cost.
 
-A block in which a large number of Tachyon stamps are aggregated into a smaller
-number of Tachyon stamps is completely verified by that smaller number of
-stamps, while still allowing every action, signature, and balance to remain
-independently valid. In the ideal case, a single proof MAY verify all Tachyon
+A block in which a large number of Tachyon stamps have been aggregated into a
+smaller number of stamps is completely verified by that smaller number of
+stamps, while every action, signature, and balance remains independently
+verifiable. In the ideal case, a single proof verifies all Tachyon
 transactions in a block.
 
-Aggregation is RECOMMENDED, not required. A miner MAY choose to include
-completely independent non-aggregated transactions.
+Aggregation is optional. Miners remain free to include non-aggregated Tachyon
+transactions; any aggregate a block does contain must be fully backed by
+adjuncts in the same block.
 
 ## Requirements
-
-(Goals, stated without conformance keywords; the normative rules that meet them live in
-the Specification.)
 
 - Reduce per-block stamp-verification cost by allowing multiple transactions' stamps to
   be merged into one aggregate stamp.
@@ -132,7 +132,7 @@ the Specification.)
 
 ## Specification
 
-The specification is organised around the 8-step aggregation lifecycle. Each step is a
+The specification is organized around the 8-step aggregation lifecycle. Each step is a
 subsection with conformance language. Two cross-cutting concerns are specified in their own
 subsections after the lifecycle and referenced from the steps that invoke them:
 [Transaction identifiers and P2P relay](#transaction-identifiers-and-p2p-relay), and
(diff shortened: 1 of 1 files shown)


book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,506 @@
...
+are summarized here non-normatively. The remaining terms are defined by this ZIP.
+
+- **Tachygram.** The `byte[32]` encoding of a field element ($\mathbb{F}_p$)
+  representing either a note nullifier or a note commitment. Consensus treats
+  nullifiers and commitments identically.
+- **Stamp.** A bundle trailer carrying a Ragu proof and the data necessary to
+  verify it, including tachygrams. The proof attests that every covered action
+  satisfies the Tachyon action rules. A bundle's stamp may be replaced by a
+  reference to a covering transaction (see Adjunct).
+- **Autonome.** A stand-alone transaction with a stamped bundle, whose proof
```
alxiong: voicing out my opposition to the term "autonome", and some nitpick on terms. 

- "autonome" is archaic, (for me, it's intimidating and hard to pronounce)
- may I suggest 3 sub-categories of tachyon bundle: **stand-alone, adjunct, aggregation** bundle
  - NOT "aggregate bundle", aggregate should describes the entire group of tachyon bundle, the merging transaction carries an "aggregation bundle"
- avoid phrasing like "stripped stamp", because it sounds like you are allowed to post a unstamped/unproven bundle. Adjuncts *are* stamped, their stamp is a pointer to an aggregate stamp. (stamp taking different shapes is categorically different than unstampped)

book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,506 @@
...
+- **Adjunct.** A transaction with a stripped bundle: its stamp has been removed
+  and replaced by a `wtxid` reference to a covering aggregate. Adjuncts retain
+  action data, action signatures, the binding signature, and `value_balance`.
+  An adjunct is valid only within a block.
+
+## Abstract
+
+Tachyon shielded transactions use a recursive proof system. Recursion allows
+many per-transaction proofs to be combined into a single proof covering all
+contributing transactions, reducing proof verification costs. This recursion
```
alxiong: - reduce verification cost *and* on-chain footprint *and* per-stamp syncing cost
- mention that proof system allow public aggregation, thus permissionless rather than designated prover only
- aggregator should be a conceptual role

book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,506 @@
...
+validation. It comprises a block-layout discipline under which miners strip
+redundant proofs, the effecting-data and authorizing-data semantics that make
+stripping safe, and P2P rules extending ZIP 239 to Tachyon's
+authorization-form malleability. All effecting data (actions, value balances,
+action signatures, and binding signatures) remains present and valid when a
+proof is stripped.
+
+## Motivation
+
+Consensus requires every proof in a block to be verified. Without aggregation,
```
alxiong: terminology nitpick again. in the same spirit of avoiding "stripping stamp" phrasing. 

consensus requires every *bundle* to be verified. Naively, every bundle carry a *proof*, consensus cost dominated by moderate proof size and proof verification. Aggregation *amortized* the cost.

I'll reframe myself from posting comments along this same line of term update.

book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,506 @@
...
+
+Aggregation is optional. Miners remain free to include non-aggregated Tachyon
+transactions; any aggregate a block does contain must be fully backed by
+adjuncts in the same block.
+
+## Requirements
+
+- Reduce per-block stamp-verification cost by allowing multiple transactions' stamps to
+  be merged into one aggregate stamp.
+- Preserve independent verifiability of every transaction's effecting data: action
```
alxiong: what about: Affecting only authorization proof mechanism: other validation including bundle balance check via binding signature, transaction authorization are remains unchanged and per-bundle.

why? "verifiability of effecting data" includes proof verification -- we only know `cv, rk` are valid by verifying the proof. so it's little inaccurate to say "independent verifiability" (even tho I understand what we are trying to say here).

book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,506 @@
...
+Transaction authors produce complete and independently verifiable transactions
+with stamped bundles (**autonomes**) and broadcast them to the mempool via the
+existing wallet RPC or P2P path.
+
+### Step 2: Aggregator observation and selection
+
+Aggregators observe transactions in mempool gossip (see
+[Transaction identifiers and P2P relay](#transaction-identifiers-and-p2p-relay)).
+
+An aggregator selects two transactions (autonomes or existing aggregates) for
```
alxiong: - selection criteria is under-specified: standalone bundles with the same anchor epoch
- "SHOULD bear disjoint tachygram sets -> "MUST", since will be rejected by consensus later otherwise.

book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,506 @@
...
+merging into a new aggregate. Selected transactions SHOULD bear disjoint
+tachygram sets.
+
+### Step 3: Witness and PCD preparation
+
+The aggregator reconstructs each selected transaction's stamp PCD from the
```
alxiong: i think after NU5, `inv` (inventory) announcement is already using `MSG_WTX` that uses `wtxid` as the identifier in the p2p network (and in mempool).
Thus iiuc, when i try to `getdata(MSG_WTX, wtxid)`, the network will give me the entire tx (affecting + auth_data). Point being, no need to "reconstruct", which is a procedure only assuming multi-stage witness fetching.

see [ZIP-239](https://zips.z.cash/zip-0239)

in another word, imo,
if I receive a request `getdata(MSG_WTX, wtxid)`, and the `wtxid` points to an adjunct bundle, which reference another `wtxid_agg`, then i will have to prepare the aggregation bundle's full tx first before responding the request. 

(agree that this is up for discussion)

book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,506 @@
...
+proof and data carried on that transaction.
+
+Merging is defined only over stamps bearing identical anchors. If the selected
+transactions bear unequal anchors, the aggregator first aligns them by fusing
+with an anchor PCD that proves the sequence from one anchor to the other.
```
alxiong: imo, unrealistic to assume a marketplace/shared pool of anchor PCD proofs that happens to fully covers `[anchor_start, anchor_end]` that an aggregator wants.

i think a cross-epoch anchor advancement proof is highly likely, because this statement is shared by all; these proofs has a canonical start anchor and end anchor (say the last-stamp in an epoch) to cover a full epoch. 
whereas in-epoch lift is most likely done by yourself, stamp-by-stamp advancement because the requested range is combinatorially many.

plus, feels simpler to just say aggregator "lift" anchors to a common anchor in the same epoch by updating the recursive proof. 
the action of "fusing" is unexplained so far, so I'd avoid it.

book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,506 @@
...
+- zero or more non-Tachyon transactions
+- zero or more Tachyon autonomes
+- zero or more Tachyon aggregates
+
+A block containing an aggregate MUST also contain, as adjuncts, every
+transaction that aggregate covers. For each such adjunct, the miner MUST strip
+the bundle's stamp and set the `tachyonAggregateId` field to the covering
+aggregate's `wtxid`.
+
+A stripped innocent (a former innocent aggregate) contributes no actions, but
```
alxiong: didn't cover "innocent" in terminology. I'd actually avoid this separation.

just say the aggregation bundle can include zero or more actions of its own, just like a stand-alone bundle.

book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,506 @@
...
+transaction is absent from the block, or bears no stamp, the validator MUST
+reject the block. A stripped bundle with no actions satisfies this check
+against any stamped transaction in the block: it contributes no action digests
+to the reconstruction in item 3, so consensus attaches no further meaning to
+its reference.
+3. **Action set commitment per stamp.** The validator MUST confirm each stamp's
+`cActionsTachyon` by reconstruction: collect the action digests of the stamped
+bundle's own actions together with those of every stripped bundle that names it
+by `tachyonAggregateId`, form the polynomial $\prod_i (X - d_i)$ over that
+combined set, and take a single Pedersen commitment of it. If the reconstructed
```
alxiong: disagree. we can save validator from re-computing it by proving it as part of the PCD proof.
the correctness of the tachyon accumulator and action accumulator should be tested in circuit (using poly-query oracle), as I explained in my writeup:

Image

This design decision is to minimize validator work, now validator only does hashing (in the anchor chain and in the action chain) without recomputing the their poly-commitment.

book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,506 @@
...
+`anchorTachyon`, a Pedersen commitment to the stamp's `vTachygrams` list of
+tachygrams, and the confirmed `cActionsTachyon`. If any proof does not verify,
+the validator MUST reject the block.
+
+### Transaction identifiers and P2P relay
+
+These semantics underpin publication (Step 5), observation (Steps 2 and 6), and stripping
+(Step 7).
+
+**Identifiers.** A Tachyon transaction is identified by `wtxid = txid || auth_digest`
```
alxiong: imo, these definition should belong to the scope of #104.

book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,506 @@
...
+as adjuncts, and the aggregate may never be mined. Further mempool retention and eviction
+policy is a local implementation concern.
+
+**Duplicate tachygrams are transaction-invalid.** Tachygram distinctness applies at the
+transaction level: a stamped bundle whose `vTachygrams` contains a duplicate tachygram is
+invalid, and a node MUST NOT accept it into the mempool or relay it. The check is a scan
+of the public list, requiring no proof verification.
+
+**Proof verification precedes relay.** A node MUST verify a stamped bundle's proof before
+accepting the transaction into its mempool or relaying it. The stamp is sufficient for
```
alxiong: beyond proof, also need other validation like balance check and auth sig verification before accepting into mempool, right? should we be explicit?

book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,506 @@
...
+this: the node assembles the proof header from the carried `cActionsTachyon`, the stamp's
+`anchorTachyon`, and a Pedersen commitment to its `vTachygrams`, and verifies the proof
+against that header. A wrong `cActionsTachyon` cannot pass, because the proof verifies
+only against the action-set commitment it actually attests to. Mempool acceptance
+therefore requires no covered transactions; confirming that a block's actions match the
```
alxiong: >Mempool acceptance therefore requires no covered transactions; 

what does this mean?

book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,506 @@
...
+admits only the stamped forms. A stripped form's `wtxid` is still meaningful at the block
+layer, where a block commits to each included transaction by `wtxid`
+([ZIP 244](https://zips.z.cash/zip-0244)); that is what lets a block name an adjunct
+unambiguously even though relay only ever carried the stamped form.
+
+### Covered-transaction identification
+
+Identifying which transactions a stamp covers is a single primitive, used by aggregators
+preparing a merge witness (Step 3) and by miners observing and composing a block (Steps 6
+and 7).
```
alxiong: do you think we should introduce a new relay message type `getxref wtxid` that the network returns any bundle with stamp containing the `wtxid` as a reference?

the p2p node build up a graph (doubly-linked) as it add verified tx and aggregated tx group into its mempool. and upon `getxref` request, it query this graph, then subroutine-call `getdata` on itself before returning.

alxiong (commented):
my focus for first pass is: 
scope, terminology, requirements (security & functional), and core design divergence from my writeup 
(since we haven't reached a consensus yet, so i will just point out these gaps for discussion)

what i intentionally under-comment on: prose, readability, ZIP citation, ZIP idiosyncrasy (as i'm unfamiliar)

book/src/zips/tachyon-aggregator.md
alxiong: sry, i take back what I suggested above. Proving action accumulator correctness for stand-alone bundle is fine and work-saving for validator, but when aggregating, it's tricky to decide what to do with per-stamp action accumulator. If the aggregation step union them (like tachygram), then the value used for `txid` is not part of the public output, defeat the purpose of proving it in circuit. If we concatenate adjunct bundle's action accumulator value in the public instance, then it's increasing the header size. 

This nudge me to think that we should shift away from polynomial-based accumulator for actions. we don't need to prove non-membership etc. Just fall back to the Merkle-tree based accumulator for actions. Because the validator has to recompute it and PCS commit is slow. (unless we pick hash-based like FRI or tensor-based like Ligero PCS which requires more implementation scope) 

in fact, probably not even MT-based, because we want to keep validator's hot memory usage low, so we would keep the hash-chain structure so that validator only keeps the chain tip rather than the whole tree. So I think we can just use simple vector hash (not Pedersen, just Blake3) on each bundle, and fold them in a hash chain.

book/src/zips/tachyon-aggregator.md
turbocrime: - dropped innocent/based naming
- dropped "stripped" terminology

deferring autonome/adjunct/aggregate naming discussion. we should have a larger conversation about that outside of this PR thread

book/src/zips/tachyon-aggregator.md
turbocrime: expanded this section

https://github.com/tachyon-zcash/tachyon/pull/148/changes/3aa6355846865be4f1de0fc12a5aae90a91500ed#diff-a7630e158ada422a785820db5f7fd3c0a5aee02a4b0dc84da327dd39cbc33fbcL107-R107

https://github.com/tachyon-zcash/tachyon/blob/3aa6355846865be4f1de0fc12a5aae90a91500ed/book/src/zips/tachyon-aggregator.md?plain=1#L103-L107

book/src/zips/tachyon-aggregator.md
turbocrime: expanded https://github.com/tachyon-zcash/tachyon/pull/148/changes/3aa6355846865be4f1de0fc12a5aae90a91500ed#diff-a7630e158ada422a785820db5f7fd3c0a5aee02a4b0dc84da327dd39cbc33fbcL122-R136

book/src/zips/tachyon-aggregator.md
turbocrime: reworded as suggested.

but conceptually, i do think `(cv, rk)` have independence as effecting data which can be confirmed independently of proof verification, via the balance check and binding signatures. the described effect itself may be correctly authorized by a proof or it may not.

book/src/zips/tachyon-aggregator.md
turbocrime: i went back and forth on this while drafting but i think you're right. corrected

book/src/zips/tachyon-aggregator.md
turbocrime: i think the sequence is a little different:

1. an aggregator that has selected an autonome already holds sufficient data, there is no further retrieval necessary.
2. an aggregator that has selected an aggregate does need the contributing actions, but it can't getdata those by wtxid, because the aggregate doesn't identify the contributing transactions.

additionally, adjunct forms are forbidden from the mempool and never relay. so no wtxid you can getdata points at an adjunct.

the present concept is that aggregators should collect and index mempool data, and simply not attempt merges they can't satisfy.

a retrieval mechanism is now detailed as open question. i would suggest we provide a by-tachygram retrieval query, because an id-based retrieval query would require an additional and large vector on aggregates.

book/src/zips/tachyon-aggregator.md
turbocrime: cross-epoch stamp advancement is presently forbidden because it disrupts relationship of spend (nf_present, nf_next) relative to the epoch of the anchor. refactoring this would change fundamental parts of the specification and seems risky/difficult.

arbitrary merge of unequal anchors is presently forbidden because we we could not accept a third input PCD about a continuous valid sequence between contributing anchors on the input PCDs. it's possible we may be able to have the aggregator *commit* to an arbitrary sequence of anchors covering the contributors, or a multiset of contributing anchors that's later verifiable as representing members of a continuous valid sequence, but this is a significant refactor.

recording both as open questions.

book/src/zips/tachyon-aggregator.md
turbocrime: dropped the terminology

book/src/zips/tachyon-aggregator.md
turbocrime: some history/context:

the original scheme deliberately published no action set commitment. consensus was required to reconstruct the commitment in order to attempt verification. every input to the reconstruction is public data on the block. a reproducible reconstruction is a strong demonstration of validity, since the proof can only *be* verified for actions actually present and available to the verifier.

during the first draft of this zip, i decided to produce pr #149 adding `cActionsTachyon` to address a shortcoming due to the variable relationship between tachygram count and action count (which may change).

reconstruction performed by aggregators before block composition was ambiguous, because it is not precisely known how many tachygrams an action relates to. confirmation failed slow, since a wrong or incomplete action set was discovered only when the expensive proof verification failed. by carrying `cActionsTachyon` in the trailer, participants verifying the proof before it appears in a block may more quickly determine correct reconstruction.

to avoid discarding the original reconstruction-as-binding logic, i specified that the value must still be confirmed by reconstruction, and not simply trusted.

turbocrime: the action set commitment is computed in the action-proving steps and confirmed in the merge step, and is an output header field. so i don't want to change the commitment scheme.

txid binds the action set commitment transitively. action digests are produced from action descriptors (cv, rk) which are bound by sighash and thus by txid.

the `cActionsTachyon` field is deliberately not bound by txid because the merge process requires it to be malleable.

if we relocate `cActionsTachyon` to the non-malleable part of the bundle, it raises many questions. validators would still face the reconstruction requirement. would need to decide:

- should the commitment represent only the bundle's own actions? then, it becomes useless for its intended purpose (fail-fast confirmation)
-  should an aggregate use it to represent all of its contributing actions? it multiplies the problem, and now a validator must essentially reconstruct the entire aggregation history correctly, which is a much more complex problem.

turbocrime: there might be some scope confusion between tachygram accumulation and action accumulation.

the tachygram set contributes to the anchor and reconstructed pcd header. in a block or in the mempool, all data is immediately available on the relevant bundle.

the action set contributes only to the reconstructed pcd header. an autonome is self-sufficient, but aggregates require data from contributing transactions. in the mempool, this may be ambiguous, which `cActionsTachyon` was implemented to improve. in a block, all data is immediately available on the block, and relationships are obvious.

at transaction authoring time and during aggregation, the relationship of any set to the anchor cannot be known. at block authoring time, when the miner is the first actor to identify anchor position, all sets already exist and must be fixed.

turbocrime: if the situation changes and outputs produce 2 tachygrams like spends, fixing the tachygrams:actions relationship as 2:1, i would prefer to simply remove `cActionsTachyon` and require total reconstruction again.

turbocrime: i agree it is possible and might be reasonable to specify `cActionsTachyon` itself to use a different commitment scheme, since its purpose is pure external coordination, and for this purpose it may be distinct from the actual action set commitment used in the proof system. i'll add this to open questions.

book/src/zips/tachyon-aggregator.md
turbocrime: relocated

book/src/zips/tachyon-aggregator.md
turbocrime: added language about this

book/src/zips/tachyon-aggregator.md
turbocrime: sorry, it's reworded.

book/src/zips/tachyon-aggregator.md
turbocrime: no mempool form names another transaction by wtxid (adjuncts naming an aggregate wtxid are presently block-only).

the association available in the mempool is tachygram matching. open questions now includes some discussion.

book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,536 @@
...
+role, the aggregator, without creating a new trust assumption.
+
+This ZIP specifies the aggregator protocol: an 8-step lifecycle from
+transaction authorization, through the mempool, to block layout and final
+validation. It comprises a block-layout discipline under which miners replace a
+covered bundle's proof with a reference to a covering transaction, the
+effecting-data and authorizing-data semantics that make stripping safe, and P2P
+rules extending ZIP 239 to Tachyon's authorization-form malleability.
+Aggregation affects only the authorization-proof mechanism: every bundle's
+effecting data (actions, value balances, action signatures, and binding
```
LucidSamuel: "effecting data" is defined in two incompatible ways (L131–135, L162–165) list it as a 4-item set including signatures. the P2P-relay section (L319–320), where the term is load-bearing for the txid-stability argument, defines it as `action_acc || value_balance`, excluding signatures (which are authorizing data).

also a likely substitution error: L164 says "action digests" where it means "actions." 
we should use one definition; signatures are authorizing, not effecting.

book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,536 @@
...
+carries its own proof, and consensus cost is dominated by proof size and proof
+verification. Aggregation amortizes that cost: it reduces the proof
+verification a validator performs, the on-chain footprint the proofs occupy,
+and the per-stamp cost of syncing the chain.
+
+The proof system permits public aggregation of already-published proofs, so the
+aggregator is a permissionless, conceptual role that any participant may take,
+not a designated prover. A block in which a large number of Tachyon stamps have
+been aggregated into a smaller number of stamps is completely verified by that
+smaller number of stamps. Aggregation affects only the authorization-proof
```
LucidSamuel: Step 8 requires public-data checks too

book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,536 @@
...
+validation.
+
+**Data availability.** Aggregation removes redundant proof bytes only. Every adjunct
+retains its action data, action signatures, binding signature, and `value_balance`;
+validators reconstruct the aggregate header from this public data. An aggregate proof
+alone is insufficient: the covered effecting data is present in the block as adjuncts,
+and the [block-validity rules](tachyon-bundle.md#block-validity) reject any block where it
+is not.
+
+**`cActionsTachyon` is confirmed, not trusted.** The action set is bound by the proof.
```
LucidSamuel: this topic sentence contradicts the rest of its own paragraph: "the action set is bound by the proof," then the next three sentences establish that it's consensus *reconstruction* that ties the header to the actual action set ("not to a value the prover supplies"), the proof only binds to the header value it was built against.

book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,536 @@
...
+fail-fast completeness check over the action-digest set, reusing the commitment
+the proof already attests to rather than adding a signed coverage manifest or a
+tachygram-origin query protocol.
+
+**Overlapping merges self-invalidate.** The disjoint-selection guidance of Step 2 is a
+SHOULD rather than a consensus rule because a violation cannot survive: merging stamps
+whose tachygram sets overlap produces a stamp bearing duplicate tachygrams, which no node
+accepts into its mempool and no valid block can contain (see
+[block validity](tachyon-bundle.md#block-validity)). The deeper reason is
+that the merge is homomorphic over the stamps' set commitments, so the overlapping stamp
```
LucidSamuel: multiset union maps to polynomial *multiplication*, not the addition a Pedersen commitment is homomorphic under, and there's no way to combine two set commitments algebraically to get the union's (the aggregator reproves it in-circuit). 

we could use the framing from #157 "multiset union, which the root-polynomial representation expresses as polynomial multiplication." then dropping "homomorphic" here and reserving it for the value commitments

book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,536 @@
...
+commits to multiply-counted multiset members that no reconstruction over a block's
+distinct transactions can reproduce. The aggregator's proving work is simply wasted, so
```
LucidSamuel: think we can drop "simply" here

book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,536 @@
...
+complete collection of action digests reproduces the action set commitment on the
+selected stamp.
+
+Aggregators therefore maintain an index of recent mempool transactions, along
+with recent consensus data and cached anchor lift proofs, and do not attempt
+merges whose contributors they cannot recover from data they hold.
+
+### Step 4: Aggregate construction
+
+Holding two stamp PCD with identical anchors and the prepared witness, the
```
LucidSamuel: ... "two stamp PCDs"

book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,536 @@
...
+  answer it; a list of contributing transactions carried on aggregates would serve only
+  the merge-input case, at the cost of a large additional vector.
+- **`cActionsTachyon` commitment scheme.** The covered-transaction identification check
+  reconstructs the carried `cActionsTachyon` field. Whether that field, which serves
+  only external coordination, uses a different commitment scheme than the in-circuit
+  action-set commitment is an open question of the
+  [bundle ZIP (#104)](https://github.com/tachyon-zcash/tachyon/issues/104).
+- **Cross-epoch lifting.** A stamp lift never crosses an epoch boundary, because
+  a spend's published nullifiers are fixed relative to its anchor's epoch.
+  Supporting them may be possible, but would change fundamental parts of the
```
LucidSamuel: supporting "cross-epoch lifts" instead of "them"

book/src/zips/tachyon-aggregator.md
```diff
@@ -6,6 +6,536 @@
...
+**Identifiers.** A Tachyon transaction is identified by `wtxid = txid || auth_digest`
+([ZIP 239](https://zips.z.cash/zip-0239), [ZIP 244](https://zips.z.cash/zip-0244)). The
+digest inputs (which fields each contribution commits to) are defined normatively by the
+[Tachyon Bundle / Aggregate Transaction Format](tachyon-bundle.md) ZIP, and the digest
+algorithm by the [ZIP 244 update](zip-244.md); this section summarizes them
+non-normatively.
+
+- Tachyon's contribution to `txid` commits only to effecting data
+  (`action_acc || value_balance`). It is stable across
+  stamping, merging, stripping, and re-stamping, so a transaction's logical identity is
```
LucidSamuel: "re-stamping" vs "restamping" (L332/340/463) we should pick one for consistency.

LucidSamuel (changes requested):
left some nits;

non blocking suggestions:
- Motivation (L147–152) says a block is "completely verified by that smaller number of proofs" Step 8 needs the public-data checks too, qualify "modulo Step 8's checks".
- define `n` in the `$O(n)$` at L403 (only unanchored asymptotic in the doc).
- Terminology entries (L98–115) are fragmented, maybe add the copula ("A tachygram is the…").
- no mempool rule for two *unrelated* txns sharing a tachygram (ordinary double-spend): maybe state it, or say it's local policy like the retention rule at L349.
- the empty-multiset `cActionsTachyon` encoding is undefined (zero-action bundles are required representable, but "the empty multiset has no commitment" vs a mandatory 32-byte point), the fix lands in #104 same empty-set gap i flagged on #157, worth resolving there and cross-referencing.

book/src/zips/tachyon-aggregator.md
turbocrime: review https://github.com/tachyon-zcash/tachyon/pull/166 for present direction of non-proof action commitments

book/src/zips/tachyon-aggregator.md
turbocrime: redefined effecting data digest https://github.com/tachyon-zcash/tachyon/pull/166

book/src/zips/tachyon-aggregator.md
turbocrime: the no longer exists as described this zip, so the commitment must always be reconstructed. review https://github.com/tachyon-zcash/tachyon/pull/166

book/src/zips/tachyon-aggregator.md
turbocrime: technically the correct pluralization would be Proofs Carrying Data, so PsCD. lol.

i think a reader would understand PCD as singular or plural, it is natural to use an acronym as either.

Issue #147, opened 2026-06-26
Stamp trailer should bear action set commitment


turbocrime:
An aggregate's stamp contains a proof for its own actions plus all covered adjuncts' actions, but no proof header. Proof verifiers must reconstruct the header. When published in a block, all material is immediately at hand. Conveniently, all of the stripped adjuncts point towards their aggregate by wtxid.

However, before block inclusion, aggregators do not gossip stripped transactions bearing wtxid labels. They only publish their own transactions with merged stamps. Participants are expected to correlate published aggregates with the autonomes they have seen before.

Mempool observers are not omniscient. Thanks to the ambiguity of action-to-tachygram relationships, it might be difficult to recognize a nearly-complete collection of adjuncts, requiring ultimate confirmation from the proof system.

**Add `tachyonActionSet` to stamp trailer.**

To provide a fast-failing indicator, a stamped bundle should bear a multiset commitment to all the actions it covers, identical to the corresponding PCD header field which presently requires unguided reconstruction.

This would allow a validator to avoid an expensive attempted verification for these purposes:

- identify an autonome.
- identify an aggregate.
- confirm a collection of adjuncts is complete.

The multiset commitment, like the rest of the strippable stamp trailer, would not be considered effecting data, but would be considered authorization data.

Publishing an unsatisfiable `tachyonActionSet` would be a self-DoS, as mempool actors would simply avoid handling your transaction.

Publishing a subset, superset, or some partial overlap `tachyonActionSet` may at worst result in consensus actors confirming the transaction does not validate.

turbocrime:
completed by #149

PR #149, opened 2026-06-26
Stamp trailer includes action set commitment 

turbocrime:
https://github.com/tachyon-zcash/tachyon/issues/147

adds a stamp action set commitment to the stamp trailer.

forbids serializing or de-serializing a zero aggregate id.


book/src/transaction-identifiers.md
```diff
@@ -21,12 +21,12 @@ $$
 \text{BLAKE2b-256}_{\text{``ZTxAuthTachyHash''}}\bigl(\\
 \quad \mathsf{vActionSigs} \,\|\, \mathsf{bindingSig} \\
 \quad\|\,\begin{cases}
-    \text{anchor}_h \,\|\, \text{anchor}_{\text{pc}} \,\|\, \text{nTachygrams} \,\|\, \text{vTachygrams} \,\|\, \text{proof} & \text{if stamped}\\
+    \text{cActionsTachyon}_{32} \,\|\, \text{anchor}_h \,\|\, \text{anchor}_{\text{pc}} \,\|\, \text{nTachygrams} \,\|\, \text{vTachygrams} \,\|\, \text{proof} & \text{if stamped}\\
```
LucidSamuel: the stamped auth_digest formula still isn’t byte-exact: the code hashes
`cActionsTachyon || anchor || vTachygrams || proof`
but the doc formula includes extra structure like `nTachygrams`. and since this PR is already touching that formula for `cActionsTachyon`, it’d be good to either make the preimage match the code or explicitly mark it as conceptual.

crates/tachyon/src/bundle/tests.rs
LucidSamuel: these cases now seem to fail through the new `ActionSetMismatch` fast-fail path, but the tests only check “some error”. worth pinning at least one of them, or using a small helper for all four, so this doesn’t silently regress back to proof failure later

```
  let err = stamped.stamp.verify(rng, &actions).unwrap_err();
  let crate::stamp::VerificationError::ActionSetMismatch = err else {
      panic!("expected ActionSetMismatch, got {err:?}");
  };
```

book/src/aggregation.md
```diff
@@ -133,6 +133,16 @@ flowchart LR
     end
 ```
 
+## Action set indicator
+
+An aggregate's stamp proves its own actions together with every covered adjunct's actions, but carries no proof header: a verifier reconstructs the header from the covered actions. Within a block that material is all at hand, since each adjunct points to its aggregate by `wtxid`. Before block inclusion it is not: aggregators publish only their merged transactions, and the ambiguity between actions and tachygrams makes a nearly-complete collection of adjuncts hard to recognize without falling back on full proof verification.
+
+To make that recognition cheap, the stamp trailer carries `cActionsTachyon`, an assistive indicator the transaction author provides: the action-set commitment the proof already attests to, mirrored onto the trailer (see [Bundles → Wire Format](./bundle.md#stamp-trailer)). It commits to the action digests of every action the stamp covers, so an observer can cheaply identify an autonome, identify an aggregate, and judge whether a collection of adjuncts is complete, without verifying the proof.
```
LucidSamuel: the doubled "identify" reads awkwardly and the use of "an autonome" here didn't give me clarity...
think we can just tighten to the indicator's actual job like "…so an observer can cheaply recognize aggregates and judge whether a collection of adjuncts is complete, without verifying the proof." or something better as such.

LucidSamuel (commented):
Clean and LGTM, the implementation, wire/auth-digest ordering, tests, and new book
framing all look consistent to me. added a couple small non-blocking comments below.

LucidSamuel (approved):
GG;

turbocrime:
in #148 i specified that stripped innocents should have an agg ref and not use zero, so i'll fold that in here

PR #150, opened 2026-07-02
chore: MSRV 1.90, lints, justfile

turbocrime:
Clippy and rustfmt config changes, justfile cleanup.

MSRV 1.85 to 1.90 matching ragu.


justfile
```diff
@@ -5,17 +5,17 @@ fmt:
...
 
 lint:
-    cargo +nightly clippy --workspace --all-targets # without features
+    cargo +nightly clippy --workspace --lib --no-default-features # no_std
     cargo +nightly clippy --workspace --all-targets --all-features
 
-test:
-    cargo test --workspace --all-features
+test *ARGS:
+    cargo test --release --workspace --all-features {{ARGS}}
```
LucidSamuel: if release-profile testing is deliberate (proof timing), can we move the semantics into the profile instead `[profile.release] overflow-checks = true` (and maybe `debug-assertions = true`) in Cargo.toml so any invocation keeps the checks? otherwise drop `--release` here; callers can pass it via `{{ARGS}}` now anyway.

Cargo.toml
```diff
@@ -206,7 +201,16 @@ arbitrary_source_item_ordering = "allow"
...
+missing_panics_doc = "allow"
 todo = "allow"
 unimplemented = "allow"
+
+# warn, deny later
 dbg_macro = "warn"
+panic = "warn"
+
+# deny, forbid later
+indexing_slicing = "deny" # use .get() or iterators. test/const exceptions via clippy.toml.
```
LucidSamuel, line 221: the test/const exceptions don't need the forbid->deny downgrade: `allow-indexing-slicing-in-tests` and `suppress-restriction-lint-in-const` suppress at emission level, so they work under `forbid` too, claude verified by flipping this back to forbid (clean) and then toggling the clippy.toml keys off (real errors appear in bundle/tests.rs, proving they're load-bearing at forbid). since deny admits silent `#[allow]` escapes where forbid was a hard error, and "deny, forbid later" was already at "later" before this PR, suggest keeping `forbid`.

LucidSamuel (commented):
Looks good, left some nits.

Cargo.toml
turbocrime: there's way more indexing use in non-test code in the upcoming PRs, and it's simpler to write and understand with indexing. this is why it's relocated.

LucidSamuel: got it, fair enough

justfile
turbocrime: all the expensive fft is in dep code. so instead cd459fa added a `[profile.dev.package."*"]` to improve dep perf in tests without using a nonstandard test profile for the in-repo code.

PR #125, opened 2026-05-28
Let `Unspent` ranges cross epoch boundaries

turbocrime:
Relocates epoch-rollover work from the spendable lineage into the `Unspent` lineage. An `Unspent` range may now span one epoch boundary, rotating its exclusion nullifier `nf_E → nf_{E+1}` and applying `Anchor::next_epoch` internally. `SpendableLift` becomes the single way a spendable advances (intra- or cross-epoch); the two spendable-side rollover steps are removed.

### Before / after

Previously a spendable crossed a boundary via `RolloverFuse`/`DelegateRolloverFuse` → `SpendableRollover` (`next_epoch`) → `SpendableEpochLift`, while the `Unspent` was intra-epoch by construction. Now the boundary work lives in one `Unspent`-lineage step, and the spendable lineage shrinks to `SpendableInit` + `SpendableLift`.

### Changes

**`pool.rs`**
- `Unspent::Data`: `(nf, start, end)` → `(start_nf, end_nf, start, end)`. Intra-epoch: `start_nf == end_nf`; cross-epoch: `nf_E → nf_{E+1}`. `encode()` +32 bytes; `SUFFIX` (6) unchanged.
- `UnspentSeed` / `EmptyBlockUnspentSeed`: emit `(nf, nf, start, end)`.
- `UnspentFuse`: check `left.end_nf == right.start_nf` (+ `left.end == right.start`); emit `(left.start_nf, right.end_nf, left.start, right.end)`.
- **New `UnspentRollover`** (seed; `Left = NullifierRolloverHeader`, witness `(start,)`): emits `(old_nf, new_nf, start, start.next_epoch(new_epoch))`. The only step invoking `Anchor::next_epoch`. Seed form so a spendable at the epoch-final anchor needs no prior-epoch segment.

**`spendable.rs`**
- `SpendableLift`: check `start_nf == spendable.nf` and `start == spendable.anchor`; emit `(end_nf, end)`.
- Removed `SpendableRollover`, `SpendableEpochLift`, `SpendableRolloverHeader`.
- Kept `RolloverFuse`/`DelegateRolloverFuse`/`NullifierRolloverHeader` (now feed `UnspentRollover`). `SpendableInit`/`SpendableHeader` unchanged (spendable stays single-nf).

**Spend path / `AnchorChain`: untouched.** Spend still happens at a single-epoch anchor and publishes `(now_nf, next_nf)` via `SpendBind`.

[...]

PR #151, opened 2026-07-02
refactor: Composable unspent fuse

turbocrime:
Base: #159

`UnspentFuse` becomes a true composition: two `Unspent` lineages sharing
a mid-epoch junction concatenate their histories, so unspent segments
compose across arbitrary contiguous ranges in any tree shape. Direct
descendant of #125 (unspent cross-epoch ranges). The witness-module and
constructor groundwork that previously padded this PR now lives in the
foundation layer; the diff here is the fuse behavior alone (four files).

- **Witness**: `()` becomes `(left, combined, right)` elapsed-sequence
  polynomials, each bound to its header commitment. The combined history
  is proven by the sentinel-concat composition of
  `enforce_shifted_combination`: the `-X^offset` monomial cancels the
  left half's sentinel, the right half's first crossing lands in the
  vacated slot, and the right half's own sentinel re-terminates the
  result.
- **Relaxed right half**: the forwards half may now carry crossings of
  its own (at the foundation layer it had to be an empty within-epoch
  extension). The junction-nullifier seam bind and anchor-adjacency
  checks are unchanged. The fused `elapsed` is the concatenation of the
  halves' histories, so it lengthens whenever the right half carries
  crossings, but the fuse itself mints no entry: the junction epoch's
  nf is either already `right_elapsed`'s head (the right half later
  crossed a boundary) or still the in-progress `nf_end` (it didn't).
  `UnspentEpochFuse` remains the only step that folds a new nullifier
  into `elapsed`.
- **`witness::unspent_fuse`** builder joins the witness module; the
  anchor-span fixture seeds one leaf per anchor step and folds them
  with `fuse_unspent_tree`, a binary fuse that derives each seam's
  step and witness from the halves' headers alone (mid-epoch concat
  when they share an epoch, boundary splice when consecutive).
- **Tests**: one composition test per fuse, running at three or four
  crossings between off-boundary, sub-block anchors (the mid-epoch
  fuse at offset 2, so the shift/cancel algebra is exercised where it
  can fail); one rejection test per reachable circuit check across
  `UnspentFuse`, `UnspentEpochFuse`, and `VerifyUnspent`, each pinned
  to its exact error string; a wide-span lift walk covering partial
  and whole epochs, including an empty block.

### DIFF
--- a/crates/tachyon/src/stamp/proof/pool.rs
+++ b/crates/tachyon/src/stamp/proof/pool.rs
@@ -361,8 +361,9 @@ impl Step for EmptyBlockUnspentSeed {
 /// adjacent anchors (`left.anchor_last == right.anchor_prev`), and agree on the
 /// junction nullifier (`left.nf_end == right.nf_start`); their histories are
 /// concatenated (`combined = left_elapsed ++ right_elapsed`). No epoch boundary
-/// is crossed, so `elapsed` gains no entry (the junction nf is already
-/// `right_elapsed`'s head). A crossing is [`UnspentEpochFuse`]'s job.
+/// is crossed, so `elapsed` gains no entry (the junction nf is
+/// `right_elapsed`'s head if right later crossed a boundary; otherwise it stays
+/// the in-progress `nf_end`). A crossing is [`UnspentEpochFuse`]'s job.
 #[derive(Debug)]
 pub struct UnspentFuse;
 
--- a/crates/tachyon/src/stamp/proof/pool.rs
+++ b/crates/tachyon/src/stamp/proof/pool.rs
@@ -355,16 +355,14 @@ impl Step for EmptyBlockUnspentSeed {
     }
 }
 
-/// Extend an accumulated [`Unspent`] within its tip epoch by one empty segment.
+/// Compose two [`Unspent`] lineages sharing a mid-epoch junction.
 ///
-/// The forwards half (`right`) carries no crossings of its own (empty
-/// `elapsed`); both halves meet inside one epoch (`right.epoch_start ==
-/// left.epoch_end`), at adjacent anchors (`left.anchor_last ==
-/// right.anchor_prev`), and agree on the junction nullifier (`left.nf_end ==
-/// right.nf_start`). The output keeps `left`'s history and span, extending only
-/// the anchor and tip; no epoch boundary is crossed, so `elapsed` is unchanged.
-/// Composing two non-empty histories is the composable-fuse step's job; a
-/// crossing is [`UnspentEpochFuse`]'s.
+/// The halves meet inside one epoch (`right.epoch_start == left.epoch_end`), at
+/// adjacent anchors (`left.anchor_last == right.anchor_prev`), and agree on the
+/// junction nullifier (`left.nf_end == right.nf_start`); their histories are
+/// concatenated (`combined = left_elapsed ++ right_elapsed`). No epoch boundary
+/// is crossed, so `elapsed` gains no entry (the junction nf is already
+/// `right_elapsed`'s head). A crossing is [`UnspentEpochFuse`]'s job.
 #[derive(Debug)]
 pub struct UnspentFuse;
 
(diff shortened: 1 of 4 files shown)


crates/tachyon/src/stamp/proof/tests.rs
```diff
@@ -932,40 +1078,84 @@ fn unspent_fuse_rejects_nonzero_forward_half() {
...
-        BlockHeight(0)..=BlockHeight(EPOCH_SIZE - 1),
+        &[nf0, nf1],
+        BlockHeight(1)..=BlockHeight(EPOCH_SIZE),
     );
-    let m_right = build_unspent_pcd(
+
+    let (fused, ()) = PROOF_SYSTEM
+        .fuse(
+            rng,
+            pool::UnspentFuse,
```
LucidSamuel, line 971: none of the three has a rejection test (grepped the error strings: zero test hits). `UnspentEpochFuse` got exactly these in #159 (`unspent_epoch_fuse_rejects_wrong_left_seq`/`_rejects_wrong_combined`), so this is a parity gap and since these are prover-side checks, deleting any of the three keeps the whole suite green.

happy to push a follow-up adding the three `UnspentFuse` rejection tests mirroring the `UnspentEpochFuse` ones (tamper left seq, tamper right seq, forge combined)

crates/tachyon/src/stamp/proof/tests.rs
```diff
@@ -932,40 +1078,84 @@ fn unspent_fuse_rejects_nonzero_forward_half() {
 
     let nf0 = Nullifier::from(Fp::random(&mut *rng));
     let nf1 = Nullifier::from(Fp::random(&mut *rng));
-    let m_left = build_unspent_pcd(
+    // Left half: epoch 0's first block only (single-epoch).
+    let left =
```
LucidSamuel: the identity degenerates to `combined = 1 + right − 1 = right`, where even dropping the left operand entirely would pass. instrumented it: all 79 `UnspentFuse` instances this test drives run at offset 0. 

i'll fold a deterministic multi-epoch-**left** case (offset ≥ 2) into the same follow-up as the rejection tests, so the shift/cancel algebra is actually exercised where it can fail.

crates/tachyon/src/stamp/proof/pool.rs
```diff
@@ -278,36 +320,49 @@ impl Step for EmptyBlockUnspentSeed {
...
+/// Compose two [`Unspent`] lineages sharing a mid-epoch junction.
 ///
-/// The forwards half crosses no boundary, both halves share the tip
-/// `present_nf` at adjacent anchors, and the output keeps `left`'s span while
-/// only extending the anchor.
+/// The halves meet inside one epoch (`right.epoch_start == left.epoch_end`), at
+/// adjacent anchors (`left.anchor_last == right.anchor_prev`), and agree on the
+/// junction nullifier (`left.nf_end == right.nf_start`); their histories are
+/// concatenated (`combined = left_elapsed ++ right_elapsed`). No epoch boundary
+/// is crossed, so `elapsed` gains no entry (the junction nf is already
```
LucidSamuel: "(the junction nf is `right_elapsed`'s head if right later crossed a boundary; otherwise it stays the in-progress `nf_end`)".

LucidSamuel (approved):
LGTM. 
the only gap is test coverage. the three witness-dependent checks that replace the deleted "empty forwards half" constant check have no rejection tests (`UnspentEpochFuse` got them in #159), and the one deterministic composition test runs at offset 0, where the sentinel-shift algebra degenerates and can't fail. 

will push a follow-up adding the three rejection tests + a deterministic offset≥2 case against this branch.

crates/tachyon/src/stamp/proof/tests.rs
turbocrime: i had a more extensive test base that seems to have been neglected in some of claude's rebase logic. i'll restore it

crates/tachyon/src/stamp/proof/tests.rs
turbocrime: i had a more extensive test base that seems to have been neglected in some of claude's rebase logic. i'll restore it

crates/tachyon/src/stamp/proof/tests.rs
turbocrime: recovered tests in the branch now

crates/tachyon/src/stamp/proof/tests.rs
turbocrime: same, recovered tests in the branch now

crates/tachyon/src/stamp/proof/pool.rs
turbocrime: corrected

PR #156, opened 2026-07-03
fix: use blake2b-256 for tx identifiers

turbocrime:
Aligns the transaction digest contributions with the ZIP 244 digest-tree
convention, checked against `orchard/src/bundle/commitments.rs` and
`librustzcash`'s `transaction/txid.rs`:

- `bundle_commitment`, `stamped_auth_digest`, and `stripped_auth_digest` are
  now BLAKE2b-256 with 32-byte outputs, matching the ZIP 244 leaves (they were
  BLAKE2b-512). The absent-bundle constants shrink to match.
- The txid contribution digests the 32-byte compressed encoding of
  `action_acc` rather than its affine coordinates. ZIP 244 digests points as
  their wire encodings; this also removes the identity-coordinates panic path.
- Auth-digest preimages remain flat field encodings with no vector counts,
  exactly the shape of the Sapling and Orchard auth leaves. Preimage shape is
  pinned by the paired effecting digest, which fixes the action count through
  `action_acc`.
- Hash-to-scalar preimages (PRF^expand, action alpha) stay BLAKE2b-512: their
  64-byte outputs feed `from_uniform_bytes` for unbiased scalar reduction, per
  protocol spec §5.4.2 and the `Zcash_ExpandSeed` convention.
- `digest/blake2b.rs` is restructured around distinct `hasher_256` /
  `hasher_512` constructors taking preimage-feeding closures, with a fully
  byte-oriented interface (tachygrams passed as encodings).
- `AggregateId` is documented as the real ZIP 239 `wtxid`:
  `txid || auth_digest`, two 32-byte ZIP 244 digests.

Book updated to match: the `authorization.md` bundle-commitment formula and
the BLAKE2b-256/512 split in `domain-separators.md`. The
`transaction-identifiers.md` formula already said BLAKE2b-256; the
implementation now agrees with it.

The `"ZTxIdTachyonHash"` / `"ZTxAuthTachyHash"` personalizations remain
placeholders pending the Tachyon amendment to ZIP 244.

### DIFF
--- a/crates/tachyon/src/bundle/mod.rs
+++ b/crates/tachyon/src/bundle/mod.rs
@@ -68,8 +68,9 @@ use core::ops;
 
 use corez::io::{self, Read, Write};
 use derive_more::{Debug, Display, Eq as TotalEq, Error, From, PartialEq};
+use ff::PrimeField as _;
 use group::GroupEncoding as _;
-use pasta_curves::{Eq, Fp, group::Curve as _};
+use pasta_curves::{Eq, Fp};
 use rand_core::{CryptoRng, RngCore};
 
 pub use crate::digest::blake2b::{AUTH_DIGEST_NO_BUNDLE, COMMIT_NO_BUNDLE};
@@ -294,17 +295,18 @@ impl Plan {
     ///
     /// This contributes to the transaction sighash.
     ///
-    /// $$ \mathsf{bundle\_commitment} = \text{BLAKE2b-512}(
-    /// \text{"ZTxIdTachyonHash"},\; \mathsf{action\_acc}_x \|
-    /// \mathsf{action\_acc}_y \| \mathsf{value\_balance}) $$
+    /// $$ \mathsf{bundle\_commitment} = \text{BLAKE2b-256}(
+    /// \text{"ZTxIdTachyonHash"},\;
+    /// \mathsf{encoding}(\mathsf{action\_acc}) \|
+    /// \mathsf{value\_balance}) $$
     ///
     /// where $\mathsf{action\_acc}$ is the polynomial commitment to the
     /// action digest multiset `∏(X - action_digest_i)` — order-independent
     /// by construction since the polynomial is invariant under root
-    /// permutation.
+    /// permutation — digested as its 32-byte point encoding.
     ///
     /// The stamp is excluded because it is stripped during aggregation.
-    pub fn commitment(&self) -> Result<[u8; 64], CommitError> {
+    pub fn commitment(&self) -> Result<[u8; 32], CommitError> {
         let digests: Vec<ActionDigest> = self
             .iter_actions(action::Plan::digest, action::Plan::digest)
             .collect::<Result<Vec<ActionDigest>, ActionDigestError>>()?;
@@ -312,7 +314,7 @@ impl Plan {
         let action_commit = ActionSetPoly::from_iter(digests).commit();
 
         Ok(blake2b::bundle_commitment(
-            &Eq::from(action_commit).to_affine(),
+            &Eq::from(action_commit).to_bytes(),
             self.value_balance()?,
         ))
     }
@@ -577,9 +579,10 @@ impl Bundle<Stamp> {
 
     /// Tachyon's contribution to the transaction `auth_digest`.
     ///
-    /// Hashes action signatures, the binding signature, and the trailer.
+    /// Hashes action signatures, the binding signature, and the stamp
+    /// trailer's field encodings.
     #[must_use]
-    pub fn auth_digest(&self) -> [u8; 64] {
+    pub fn auth_digest(&self) -> [u8; 32] {
         let action_sigs: Vec<[u8; 64]> = self.actions.iter().map(|act| act.sig.into()).collect();
         let binding_sig: [u8; 64] = self.binding_sig.into();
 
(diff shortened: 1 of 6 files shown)


crates/tachyon/src/digest/blake2b.rs
```diff
@@ -80,73 +106,61 @@ const AUTH_DIGEST_PERSONALIZATION: &[u8; 16] = b"ZTxAuthTachyHash";
...
+    })
 }
 
 /// A stamped bundle's contribution to the transaction auth_digest.
 ///
-/// Hashes action signatures, the binding signature, and the trailer.
+/// Hashes action signatures, the binding signature, and the stamp trailer's
+/// field encodings: the action-set commitment, the anchor, each tachygram,
+/// and the proof. Vector counts are not part of the preimage, matching the
+/// ZIP 244 auth-digest leaves.
```
LucidSamuel, line 125: assumption clarification: in ZIP 244 every variable-length auth field is pinned by id-side data (nActionsOrchard fixes both sig count and proof length). but here the txid side commits only to action_acc || value_balance ergo the tachygram count is bound nowhere except by length arithmetic: it falls out only because the proof is compile-time fixed-size (PROOF_SIZE_COMPRESSED = 23_000) and the action count is pinned by action_acc's monic leading term and neither invariant is stated in the digest module.

crates/tachyon/src/bundle/mod.rs
```diff
@@ -602,7 +605,11 @@ impl Bundle<Stamp> {
             &binding_sig,
             &trailer.0,
             &trailer.1,
-            &trailer.2,
+            &trailer
+                .2
+                .iter()
+                .map(|&tg| tg.to_repr())
```
LucidSamuel: non blocking: tachygrams get collected twice now `Vec<Fp>` in the trailer block above, then `Vec<[u8; 32]>` here. one fused pass in the trailer block `self.stamp.tachygrams.iter().map(|&tg| Fp::from(tg).to_repr()).collect::<Vec<[u8; 32]>>()` would keep the byte-oriented digest interface with a single allocation.

crates/tachyon/src/bundle/mod.rs
```diff
@@ -602,7 +605,11 @@ impl Bundle<Stamp> {
             &binding_sig,
             &trailer.0,
```
LucidSamuel, line 605: trailer block above re-derives the same field encodings `Stamp::write` produces, and nothing ties the two sites together so no test would catch drift, and a newly added stamp field would compile cleanly while silently escaping the auth digest. destructuring the stamp there (`let Stamp { action_set, anchor, tachygrams, proof } = ..`) would make a new field a compile error.

LucidSamuel (commented):
GG;
One comment I'd like answered before #157 makes these preimages normative, the flat auth-digest preimage is injective today, but only because of two invariants stated nowhere (fixed proof size; action count pinned by the txid via `action_acc`). see the inline comment on `stamped_auth_digest`.

The rest are non-blocking nits.

crates/tachyon/src/digest/blake2b.rs
turbocrime: i'm not sure what your point/suggestion is here.

specifically, when digesting components of the sapling and orchard bundles, no digest includes a vector length for a collection. the digest state simply consumes each member of the collection.

crates/tachyon/src/bundle/mod.rs
turbocrime: serialization and auth_digest are two distinct paths, and there's no cross-method scope here.

we may actually want to place elements in the stamp which are not digested, so i don't think this is a problem.

to improve semantics, and alignment with ZIP-244 which specifies sub-bundle digests, it might be good to provide stamp serialization and digest methods on the Sealed trait. but i'm going to deem that out of scope for this PR.

crates/tachyon/src/bundle/mod.rs
turbocrime: commit f1953fe6cbd34802e18e73fe6b581fe2648ea02d now avoids the intermediate alloc

crates/tachyon/src/digest/blake2b.rs
LucidSamuel: oh I was not suggesting we add a length. the point I was trying to put across was why that's safe there but only incidentally safe here.

in orchard the flat no-length digest is unambiguous because nActionsOrchard (committed id-side) pins every variable-length field's boundary so the lengths are omitted from the digest but not free, the id tree fixes them.

here the preimage ends `... || tachygrams || proof` the tachygram count isn't pinned anywhere on the txid side (bundle_commitment is action_acc || value_balance, no tachygram data) so the only reason the tachygrams|proof boundary is recoverable is that proof is a compile-time constant (PROOF_SIZE_COMPRESSED) ergo injectivity holds, but by an incidental invariant, not a committed one which isn't stated in the digest module.

~ not asking for a code change, just a comment recording the two load-bearing assumptions which is like the fixed-size question you raised with on #157. if a bulletproof final compression makes the proof variable-size, `tachygrams || proof` becomes ambiguous and two bundles collide the same wtxid so "fixed-size" is doing real work in the digest, not just the wire format.

LucidSamuel (approved):

crates/tachyon/src/digest/blake2b.rs
turbocrime: okay, thanks. this is good to call out and now f1e63266 describes this in the docstring.

we should probably specify more sub-component digests like ZIP-244 to explicitly manage properties like this.

Issue #105, opened 2026-05-24
ZIP: Tachyon Accumulator / Hash Chain


TalDerei:
**Book entry:** [`book/src/zips/tachyon-accumulator.md`](https://tachyon.z.cash/_book/zips/tachyon-accumulator.html)

## Summary

The accumulator hierarchy: at the tachystamp-level, we have a tachygram vector commitment (pedersen commitment to a multiset polynomial whose roots are poseidon hashes). At the block-level, we have a block accumulator (poseidon hash-chain that absorbs the coordinates of each stamp's pedersen commitments).

## Classification

| Field    | Value      |
| -------- | ---------- |
| Role     | Additive   |
| Category | Consensus  |

## Milestones

- [ ] Draft outline in book with design considerations
- [ ] Submitted as PR to `zcash/zips`
- [ ] Status → Draft
- [ ] Status → Proposed
- [ ] Implemented in `librustzcash` / `zebra`
- [ ] Status → Implemented
- [ ] Status → Final (post-activation)

### Zebra Implementation

- [ ] Implementation drafted

Issue #103, opened 2026-05-24
ZIP: Tachyon Shielded Protocol


TalDerei:
**Book entry:** [`book/src/zips/tachyon-shielded-protocol.md`](https://tachyon.z.cash/_book/zips/tachyon-shielded-protocol.html)

## Summary

This defines a new shielded pool alongside a simplified key structure. More specifically, the specification will define: note structure, key derivation, note commitments and nullifiers, value commitments, actions, tachygrams, spend authorization signatures, binding signatures, consensus rules (bounded nullifier sets and nullifier pruning, anchors, etc.), authorization.

The pool will cross reference Ragu / Tachyon books (like the [Orchard pool](https://zips.z.cash/zip-0224) references [Halo2](https://zcash.github.io/halo2/)). The ZIP only includes things that touch consensus, not the full proving system.

## Classification

| Field    | Value      |
| -------- | ---------- |
| Role     | Additive   |
| Category | Consensus  |

## Milestones

- [ ] Draft outline in book with design considerations
- [ ] Submitted as PR to `zcash/zips`
- [ ] Status → Draft
- [ ] Status → Proposed
- [ ] Implemented in `librustzcash` / `zebra`
- [ ] Status → Implemented
- [ ] Status → Final (post-activation)

### Zebra Implementation

- [ ] Implementation drafted

PR #157, opened 2026-07-03
Bundle / aggregate transaction format ZIP

turbocrime:
Describes behavior after #156 

Adds the ZIP draft page for the Tachyon Bundle / Aggregate Transaction Format
#104, the Tachyon analogue of ZIP 225. 

The draft specifies:

- The three-state wire format (`tachyonBundleState` 0x00 / 0x01 / 0x02) with
  field tables for the bundle body, stamp trailer, and stripped trailer,
  defined as a new section of the existing transaction format in the manner of
  ZIP 225. ZIP-248 registration is an open question, not a dependency.
- Canonical encodings: minimal compactsize with the 0x02000000 bound, point
  and field-element canonicity, opaque signatures at parse time, and the
  fixed-size proof field.
- The definitions the aggregator ZIP (#106) delegates here, made normative:
  the stamp, the Poseidon action digest (`Tachyon-ActionDg` domain over the
  cv/rk coordinates), and the root-polynomial set commitments.
- Value balance, the validator-derived bvk, and the binding- and
  action-signature rules over the transaction sighash.
- The transaction digest contribution inputs (BLAKE2b-256, flat field
  encodings; personalizations placeholder pending the ZIP 244 amendment).
- A consolidated bundle-validity rule list, with an explicit paragraph on what
  is owned elsewhere (#103 circuit statements and epoch rules, #105 anchor
  membership, #106 block validation and relay).

Design Considerations carries the open questions: ZIP-248 migration, the
digest-amendment home, the unfrozen PROOF_SIZE constant, zero-action bundles,
per-stamp duplicate-tachygram rule ownership (stated here, cited by #106, not
yet enforced in code), the missing value-balance range bound, and count caps.

Please provide suggestions and amendments.


book/src/zips/tachyon-bundle.md
```diff
@@ -6,6 +6,586 @@
 
 ## I. Dependencies
 
+- [Tachyon Shielded Protocol (#103)](https://github.com/tachyon-zcash/tachyon/issues/103)
+  defines the Tachyon pool, action semantics, tachygrams, keys, the statement the
+  stamp proof attests to, and the epoch-level tachygram rules.
+- [Tachyon Accumulator / Hash Chain (#105)](https://github.com/tachyon-zcash/tachyon/issues/105)
+  defines the anchor (Poseidon hash-chain state) that `anchorTachyon` references, and the
+  consensus anchor-membership rule.
+- [Tachyon Aggregator Protocol (#106)](https://github.com/tachyon-zcash/tachyon/issues/106)
```
alxiong: isn't this circular dep?

book/src/zips/tachyon-bundle.md
```diff
@@ -6,6 +6,586 @@
...
+  a `(bundleType, bundleVariant)` TLV entry (see [the ZIP-248 entry](zip-248.md) and the
+  registration draft in
+  [tachyon-zcash/zips#4](https://github.com/tachyon-zcash/zips/pull/4)). Under TLV
+  framing, absence of the entry encodes "no bundle", making the `0x00` state byte
+  redundant there.
+- **Transaction-digest amendment.** The `"ZTxIdTachyonHash"` and `"ZTxAuthTachyHash"`
+  personalizations and the placement of the Tachyon branches in the ZIP 244 digest trees
+  await an amendment to ZIP 244, which may be a separate update ZIP or may fold into one
+  of the general Tachyon ZIPs.
+- **Proof-size constant.** `proofTachyon` is a fixed-size field whose length is a
```
alxiong: if we use bulletproof for final compression, then won't be a constant-size. even though the step circuit size is capped, the polynomial-query oracle would allow larger degree polynomials, thus will be a variable-size proof imo.

maybe dynamic `bytes` field will work.

book/src/zips/tachyon-bundle.md
```diff
@@ -6,6 +6,586 @@
...
+### Action signatures
+
+Each action signature in `vActionSigsTachyon` MUST be a valid RedPallas signature
+(SpendAuth basepoint, for spends and outputs alike) over the transaction sighash
+under the corresponding action's $\mathsf{rk}$. The sighash itself is a
+transaction-level digest, computed as specified by
+[ZIP 244](https://zips.z.cash/zip-0244) as amended for Tachyon; all of a bundle's
+signatures sign the same sighash.
+
+### Transaction digest contributions
```
alxiong, line 336: in my writeup, i also proposed a `da_digest` for in-band memo:

Image

book/src/zips/tachyon-bundle.md
```diff
@@ -6,6 +6,586 @@
...
+deterministic functions of data published on the transaction, and carry no
+information beyond it.
+
+### Value balance and the binding signature
+
+`valueBalanceTachyon` asserts the net value leaving the Tachyon pool in this
+transaction:
+
+$$ \mathsf{valueBalanceTachyon} = \sum_{i \in \mathrm{spends}} v_i \;-
+\sum_{j \in \mathrm{outputs}} v_j $$
```
alxiong: a minor inconsistency. if we assume `v_i` for spend and output are both positive, then `cv = [v]G` is inaccurate, we should say `cv = [-v]G` for spend.

book/src/zips/tachyon-bundle.md
```diff
@@ -6,6 +6,586 @@
...
+  the field modulus.
+- `cActionsTachyon` is a 32-byte compressed Vesta point (the set-commitment group is
+  Vesta, since polynomial coefficients are Pallas base field elements, which are
+  Vesta scalars). A parser MUST reject an invalid point encoding.
+- Signatures (`vActionSigsTachyon`, `bindingSigTachyon`) are opaque 64-byte strings
+  at parse time; their validity is a verification-time property.
+- `proofTachyon` is exactly `PROOF_SIZE` bytes and MUST decode as a Ragu proof. The
+  proof encoding is defined by the Ragu proof system.
+
+### Stamp trailer
```
alxiong: in my writeup, i also have the tachygram accumulator value and epoch $e$ in the stamp. 
mine is: `( { tg_i​ }, acc^tg, anchor, e)`

quoting:
> Why anchor per-stamp rather than per-block, when the block is the unit of consensus finality? The primary justification is that it minimizes validator work, in alignment with our [philosophy](#philosophy). Each stamp already ships a acctg whose correctness is [batch-verified in circuit](#acc-correct), so a validator merely hashes it into the chain. A per-block anchor would instead force every validator to rebuild a block-wide accumulator from scratch: re-accumulating every tachygram in the block, interpolating the product polynomial, and committing it which involves an expensive multi-scalar multiplication for some PCS choices.

book/src/zips/tachyon-bundle.md
```diff
@@ -6,6 +6,586 @@
...
+| Name                  | Format               | Description                              |
+| --------------------- | -------------------- | ---------------------------------------- |
+| `cActionsTachyon`     | 32 bytes             | action set for this proof                |
+| `anchorTachyon`       | 32 bytes             | pool state reference                     |
+| `nTachygrams`         | compactsize          | number of tachygrams                     |
+| `vTachygrams`         | 32 * nTachygrams     | tachygrams for this proof                |
+| `proofTachyon`        | PROOF_SIZE blob      | serialized proof of fixed size           |
+
+`cActionsTachyon` is the action-set commitment (see
+[Set commitments](#set-commitments)) over every action the stamp covers: the bundle's
```
alxiong: open discussion whether we should use non-PCS commitment for this. (as I asked in Slack)

book/src/zips/tachyon-bundle.md
```diff
@@ -6,6 +6,586 @@
...
+
+`anchorTachyon` references the pool state the proof is valid against; its semantics
+and the anchor-membership rule are specified by the
+[Tachyon Accumulator / Hash Chain](tachyon-accumulator.md) ZIP.
+
+`vTachygrams` publishes the stamp's tachygram multiset for data availability. Which
+tachygrams an action contributes is specified by the
+[Tachyon Shielded Protocol](tachyon-shielded-protocol.md) ZIP; this ZIP imposes no
+relation between `nTachygrams` and `nActionsTachyon`, and an aggregate's stamp
+carries tachygrams for actions that are not its own. The tachygrams within one
```
alxiong: we should enforce `nTachygrams = 2 * nActionsTachyon`

quoting:
> To keep spend and output actions indistinguishable, we further require each output action to publish a random dummy tachygram alongside its note commitment.
> ...
> The accumulator collects two tachygrams per action: `(nf_e​,nf_e+1​)` for a spend, the note commitment and a zero-value dummy `(cm,tg⊥​)` for an output.

book/src/zips/tachyon-bundle.md
```diff
@@ -6,6 +6,586 @@
...
+  [Aggregator Protocol ZIP (#106)](https://github.com/tachyon-zcash/tachyon/issues/106)
+  citing it (matching that draft's open question on rule ownership). The rule is stated
+  here but not yet enforced in the reference implementation's parse or verify paths.
+- **Block-validation ownership.** This draft cites proof verification and block-level
+  validity to the
+  [Aggregator Protocol ZIP (#106)](https://github.com/tachyon-zcash/tachyon/issues/106),
+  their current home. That draft's own open questions contemplate relocating its
+  consensus rules (possibly here) or splitting into network and consensus ZIPs; the
+  citations follow wherever those rules land.
+- **Value-balance range.** No `MAX_MONEY` bound on `valueBalanceTachyon` is specified or
```
alxiong: why not a bound? doesn't it cause an overflow danger without it?

book/src/zips/tachyon-bundle.md
```diff
@@ -6,6 +6,586 @@
...
+[Tachyon Aggregator Protocol](tachyon-aggregator.md) ZIP; `txid`, `auth_digest`, and
+`wtxid` are defined by [ZIP 244](https://zips.z.cash/zip-0244) and
+[ZIP 239](https://zips.z.cash/zip-0239). They are summarized here non-normatively. The
+remaining terms are defined by this ZIP.
+
+- **Tachygram.** The `byte[32]` encoding of a field element ($\mathbb{F}_p$)
+  representing either a note nullifier or a note commitment. Consensus treats
+  nullifiers and commitments identically.
+- **Anchor.** A Poseidon hash-chain state referencing the Tachyon pool at a
+  specific block.
```
alxiong: > .. at specific block

in my writeup, i specifically clarify that anchor is per-stamp, not per-block:

> The chain therefore advances at sub-block, above-transaction granularity: in a block of transactions with all standalone Tachyon bundles it ticks once per transaction. A stamp's anchor field may reference any node value sttg in the chain, anchoring at a historical snapshot of the pool state.

Should we update the language here?

book/src/zips/tachyon-bundle.md
```diff
@@ -6,6 +6,586 @@
...
+
+- A transaction's `txid` contribution is invariant across stamping, merging,
+  stripping, and re-stamping.
+- Stamped and stripped forms carry identical effecting data; only the trailer
+  differs.
+- Every field a validator needs for signature and balance verification is present in
+  both states.
+- Encodings are canonical: each parsed bundle has exactly one serialization.
+- The proof field has a fixed size, so parsing requires no untrusted length.
+- Bundles with no actions are representable (innocent aggregates and their stripped
```
alxiong: open discussion for "innocent aggregate", why the necessity to give this category a name, we are not treating them or validating them any differently, right?

book/src/zips/tachyon-bundle.md
```diff
@@ -6,6 +6,579 @@
...
+Each action has an action digest, a Poseidon hash of its
+$(\mathsf{cv}, \mathsf{rk})$ pair. Let $(\mathsf{cv}_x, \mathsf{cv}_y)$ and
+$(\mathsf{rk}_x, \mathsf{rk}_y)$ be the affine Pallas coordinates of the action's
+decompressed $\mathsf{cv}$ and $\mathsf{rk}$. Then
+
+$$ d = \mathrm{Poseidon}\bigl(\mathsf{dom},\ \mathsf{cv}_x,\ \mathsf{cv}_y,\
+\mathsf{rk}_x,\ \mathsf{rk}_y\bigr) $$
+
+where $\mathsf{dom}$ is the Pallas base field element whose integer value is the
+little-endian interpretation of the 16-byte ASCII string `Tachyon-ActionDg`.
```
LucidSamuel: the Poseidon instance is neither specified nor deferred. the action digest is consensus-critical (since it feeds action_acc, hence the txid), yet "Poseidon" is never pinned, no width, rounds, constants, or sponge mode, and no deferral sentence. the neighboring set-commitment section shows the fix pattern as it explicitly defers its generators to "the polynomial-commitment generators fixed by the Ragu proof system" (line 317). ergo one sentence deferring the Poseidon instance to ragu's PoseidonFp closes a txid-splitting gap.

book/src/authorization.md
```diff
@@ -75,7 +75,7 @@ The bundle commitment is a digest of the bundle's effect.
 
 $$ d_i = \text{Poseidon}_\text{Tachyon-ActionDg}(\mathsf{cv}_i \| \mathsf{rk}_i) $$
```
LucidSamuel, line 76: the book's action-digest formula contradicts the ZIP, and the ZIP matches the code. authorization.md says `Poseidon_Tachyon-ActionDg(cv ‖ rk)`; the ZIP and digest/poseidon.rs absorb five field elements: (dom, cv_x, cv_y, rk_x, rk_y) with the domain as a field element. but the mermaid diagrams (lines 282/319/355) repeat the stale form, and line 150's "from ordered action digests" also misleads as the commitment is permutation-invariant (the ZIP's line 487 rationale). since the stack already edits authorization.md, fixing it here likewise makes sense.

book/src/zips/tachyon-bundle.md
```diff
@@ -6,6 +6,579 @@
...
+1. `tachyonBundleState` MUST be `0x00`, `0x01`, or `0x02`.
+2. Every compactSize MUST be minimally encoded and MUST NOT exceed `0x02000000`.
+3. Every point and field-element encoding MUST be canonical, as specified in
+   [Canonical encodings](#canonical-encodings).
+4. An action's $\mathsf{cv}$ and $\mathsf{rk}$ MUST NOT be the identity point.
+5. Every action signature MUST verify over the transaction sighash under its
+   action's $\mathsf{rk}$.
+6. `bindingSigTachyon` MUST verify over the transaction sighash under the derived
+   $\mathsf{bvk}$.
+7. A stamped bundle's `proofTachyon` MUST be exactly `PROOF_SIZE` bytes and a valid
```
LucidSamuel: rule 7 is purely syntactic, and every mention of actual proof *verification* delegates to the Aggregator Protocol ZIP, which is Category "Network". as drafted, also no Consensus-category document anywhere requires any ragu proof to verify. deferring per-transaction verification is right (aggregates verify at block level after stripping), but the Block-validation-ownership open question (lines 72–77) should maybe name the category mismatch explicitly as the consensus rule set isn't complete until #106's consensus half lands in a Consensus-category home (there, or here).

book/src/zips/tachyon-bundle.md
```diff
@@ -6,6 +6,579 @@
...
+
+where $\mathsf{dom}$ is the Pallas base field element whose integer value is the
+little-endian interpretation of the 16-byte ASCII string `Tachyon-ActionDg`.
+
+If an action's $\mathsf{cv}$ or $\mathsf{rk}$ is the identity point, it has no
+affine coordinates, the digest is undefined, and the transaction is invalid.
+
+### Set commitments
+
+Multisets of field elements are committed as polynomials with the members as roots.
```
LucidSamuel: Coefficient-to-generator ordering (the code is ascending, constant term at g₀) and the empty-set value are unspecified and the empty case is reachable, zero-action bundles are wire-valid and line 433 invokes an "encoded empty-set commitment" whose value (1·g₀, from A(X)=1) isn't defined. 

possible Fix: "the n+1 coefficients ordered by ascending degree, constant term first; the empty multiset commits to the constant polynomial 1."

book/src/zips/tachyon-bundle.md
```diff
@@ -6,6 +6,579 @@
...
+### Placement and bundle states
+
+The Tachyon bundle is a contiguous section of the transaction encoding, added by a
+Tachyon network upgrade. The first byte of the section, `tachyonBundleState`,
+selects the bundle state:
+
+| value         | state       | bundle contents                       |
+| ------------- | ----------- | ------------------------------------- |
+| `0b0000_0000` | non-tachyon | no bundle                             |
+| `0b0000_0001` | stamped     | bundle with anchor, tachygrams, proof |
```
LucidSamuel: missing `cActionsTachyon` ergo it disagrees with the Terminology "Stamp" entry, this page's own trailer table (line 330), and bundle.md's state table (which got this cell in #149).

book/src/zips/tachyon-bundle.md
LucidSamuel: agreed it needs a bound  think it we add it as Bundle validity rule 9 now, "`valueBalanceTachyon` MUST be in the range −MAX_MONEY to MAX_MONEY inclusive", which matches ZIP 225's bound on `valueBalanceOrchard`. this ZIP already owns the single-bundle rules, so ownership is naturally here rather than the transaction layer.

book/src/zips/tachyon-bundle.md
```diff
@@ -6,6 +6,579 @@
...
+  framing, absence of the entry encodes "no bundle", making the `0x00` state byte
+  redundant there.
+- **Transaction-digest amendment.** The `"ZTxIdTachyonHash"` and `"ZTxAuthTachyHash"`
+  personalizations and the placement of the Tachyon branches in the ZIP 244 digest trees
+  await an amendment to ZIP 244, which may be a separate update ZIP or may fold into one
+  of the general Tachyon ZIPs.
+- **Proof-size constant.** `proofTachyon` is a fixed-size field whose length is a
+  constant of the Ragu proof system. The constant is not yet numerically frozen, so this
+  draft names it symbolically as `PROOF_SIZE`.
+- **Zero-action bundles.** Stamped bundles with no actions (innocent aggregates) and
```
LucidSamuel: think we can add instead "a bundle with no actions MUST have `valueBalanceTachyon = 0`" to Bundle validity. the binding-sig infeasibility argument is sound but leaves a consensus property implicit and it's the analogue of v5, where `valueBalanceOrchard` is simply absent when `nActionsOrchard = 0`.

LucidSamuel (commented):
WIP: checked the draft field-by-field against the reference implementation and the rest of the book. the wire tables, trailer layouts, and digest preimages all match `Stamp::read`/`write` and `digest/blake2b.rs`. where the ZIP and the book chapters disagree (action-digest formula in authorization.md), the ZIP is the one that matches the code.

Inline comments cover the rest.

book/src/zips/tachyon-bundle.md
turbocrime: @ebfull previously specified that i should use fixed-size as a working assumption. if this has changed, what are the details?

book/src/zips/tachyon-bundle.md
turbocrime: the name was invented because previous conversations were difficult without a precise term. practically, they have some unique properties.

i think the strangest property is probably understated. it will be interesting for auditors.

essentially, once an innocent bundle is stripped, it contributes no information to an aggregate proof, and its aggregate reference could be arbitrarily assigned.

https://github.com/turbocrime/tachyon/blob/280875bbf3ea53391ad2b17a963de2c9c1773eee/book/src/zips/tachyon-aggregator.md?plain=1#L474-L476

book/src/zips/tachyon-bundle.md
turbocrime: dropped the ref

book/src/zips/tachyon-bundle.md
turbocrime: let's defer this. can we issue and PR the feature?

turbocrime: https://github.com/tachyon-zcash/tachyon/issues/163

book/src/zips/tachyon-bundle.md
turbocrime: ok. also refining the spec ref to orchard since it is the same

book/src/zips/tachyon-bundle.md
turbocrime: in any block, all epoch values on any stamp would be either the present epoch or the last epoch, so i don't think it's useful to include.

turbocrime: any tachygram accumulator value would require confirmation by reconstruction, so i don't think that's useful to include either.

we could specify an alternate digest like `cActionsTachyon` may adopt, but i think the benefit is minimal, since the set must still be constructed for verifying the proof.

https://github.com/tachyon-zcash/tachyon/issues/162 is open for more structured digests which could be relevant or satisfy some of these things.

book/src/zips/tachyon-bundle.md
turbocrime: https://github.com/tachyon-zcash/tachyon/issues/162 as relevant

book/src/zips/tachyon-bundle.md
turbocrime: https://github.com/tachyon-zcash/tachyon/issues/164 issued

book/src/zips/tachyon-bundle.md
turbocrime: prose updated, issued as https://github.com/tachyon-zcash/tachyon/issues/165

book/src/zips/tachyon-bundle.md
turbocrime: reworded. granularity spec belongs to the accumulator zip.

the chain does advance per-stamp, but consensus is specified to only *acknowledge* end-of-block states as valid anchors.

book/src/zips/tachyon-bundle.md
turbocrime: this is updated

book/src/zips/tachyon-bundle.md
turbocrime: issued as https://github.com/tachyon-zcash/tachyon/issues/165

book/src/zips/tachyon-bundle.md
turbocrime: i've removed the "innocent" term from the present draft

book/src/authorization.md
turbocrime: consider authorization.md out of scope for the PR, the diff there is removed

book/src/zips/tachyon-bundle.md
turbocrime: i think it was incorrect to describe an empty action set as possible, so this is reworded now

book/src/zips/tachyon-bundle.md
turbocrime: i think some concerns need to be relocated to the tachyon shielded protocol zip

book/src/zips/tachyon-bundle.md
turbocrime: the digest contents and exact details are still under discussion and development so not all of this is actually fixed yet.

book/src/zips/tachyon-bundle.md
turbocrime: the proof is now digested independently before input into the bundle digest https://github.com/tachyon-zcash/tachyon/pull/166

book/src/zips/tachyon-bundle.md
turbocrime: there is now a specified non-pcs commit for actions vector at the bundle and stamp level https://github.com/tachyon-zcash/tachyon/pull/166

book/src/zips/tachyon-bundle.md
LucidSamuel: the independent proof digest in #166 helps, but note it digests `proof.serialize()` (the re-serialization), not the received wire bytes, so the padding-canonicality issue raised (unchecked tail at parse, uncommitted by the digest) is unaffected by that change and still stands under either the fixed- or variable-size outcome. 
flagging so the proof-size decision here accounts for it (plus a self-reminder).

book/src/zips/tachyon-bundle.md
```diff
@@ -6,6 +6,646 @@
...
 ## I. Dependencies
 
+- [Tachyon Shielded Protocol (#103)](https://github.com/tachyon-zcash/tachyon/issues/103)
+  defines the Tachyon pool, action semantics, tachygrams, keys, and the statement the
+  stamp proof attests to.
+- [Tachyon Accumulator / Hash Chain (#105)](https://github.com/tachyon-zcash/tachyon/issues/105)
+  defines the anchor (Poseidon hash-chain state) that `anchorTachyon` references, the
+  consensus anchor-membership rule, and the epoch window with its duplicate-tachygram
+  rule.
+- [ZIP 225](https://zips.z.cash/zip-0225) is the v5 transaction-format precedent this
```
S1nus: tachyon will be v7, not v6 (because v6 is the nu6.3 ironwood upgrade)

S1nus: we will need a transaciton V7 ZIP as well

PR #160, opened 2026-07-04
refactor: mimc ggm

turbocrime:
GGM tree becomes arity 64 and uses 128-round MiMC.

The leaf step emits a single coverage-queryable `NullifierDerivation` header.

## The derivation

A note's per-epoch nullifiers `GGM(mk, e)` are the outputs of a 64-ary tree of
key schedules. Each node's schedule keys a 128-round MiMC cipher
`TachyonP5R128`; its 64 whitened outputs are its children's schedules, and a
depth-2 node's outputs are the nullifiers for a block of 64 consecutive epochs.

- `NfMasterSeed` derives one part of the master key from `(psi, nk)` after
  checking `note.pk == pak.derive_payment_key()` (pinning `nk`, and through it
  `cm`). The master key is split across two parts, each seeded independently.
- `NfMasterStep` assembles `mk` and proves the first child schedule out of it.
- `NfPrefixStep` descends one internal level, binding the parent schedule to its
  commitment before use and accumulating a free 6-bit chunk into the node index.
- `NullifierDerivationStep` expands a depth-2 node under a dedicated leaf domain
  into its 64 epoch nullifiers.

Each expansion proves one `64 × ROUNDS` keyed-cipher trace as a single
committed polynomial, over three generic vanishing relations (round-0 boundary,
per-round recurrence, and a decimation binding the output). The free chunk at
each level is pinned into the header index, so which leaf a walk reaches is
determined even though every step's chunk is a free witness.

## Coverage-queryable ranges

`NullifierDerivation = (cm, epoch_start, epoch_end, seq_commit)`. `seq_commit`
commits a single coefficient-form, sentinel-terminated sequence `q` over
`[epoch_start, epoch_end)` with `coeff[e − epoch_start] = nf_e`. The header
carries no boundary nullifiers: a consumer reads whatever single leaf or
sub-range it needs by coverage.

- `NullifierFuse` shift-concatenates any two contiguous derivations (same `cm`,
  `right.start == left.end`) into a wider one, up to the polynomial-size budget.
- Consumers extract by coverage via shifted-combination decompositions of `q`;
  the sentinels pin each part's length so, with header-fixed offsets, the
  decomposition is unique. Leaves are read as genuine degree-0 openings, never
  free-witness monomials.

## Homomorphic eval-to-coeff bind

`NullifierDerivationStep` certifies the 64 leaf outputs in evaluation form as
`B` (`B(ζ^p) = nf_p` over the order-64 subgroup), then binds them to the
published coefficient-form `q` homomorphically instead of reading `B` at 64
points. A running-sum accumulator `A` carries the geometric total
`A(1) = Σ_p B(ζ^p)·β^p` at a challenge `β` derived over both commitments;
discharging `q`'s sentinel, `q(β) − β^64 == A(1)` forces every coefficient of
`q` to the genuine leaf by Schwartz-Zippel. The accumulator is verified by a
masked Horner recurrence with a degree-0 quotient (no splits).

## Split spend path

- `SpendBind` (`SpendableHeader × NullifierDerivation → SpendHeader`) binds the
  epoch nullifier pair `(present_nf, nf_next)` by coverage against a covering
  derivation, tied to the lineage by `deriv.cm == spendable.cm`. It needs no
  note witness.
- `SpendStamp` (`SpendHeader → StampHeader`) proves the action `(cv, rk)` from
  the note, mirroring `OutputStamp`, publishing the pair read off the header.

Value is pinned two independent ways: `note.commitment() == cm` (SpendStamp) and
`deriv.cm == spendable.cm` (SpendBind). Publishing both the present and next
nullifiers lets consensus apply the spend across an epoch transition that may
fall between proof construction and inclusion; the note's age never becomes
public.

Other consumers follow the same coverage model: `SpendableInit` confirms the
creation-epoch leaf by a length-1 slice; `UnspentBind` binds a sync-built
`Unspent` by extracting its `elapsed ++ tip` sub-range from the derivation.


turbocrime:
i think we're not using this

Issue #162, opened 2026-07-08
Specify more structured digests


turbocrime:
relevant to https://github.com/tachyon-zcash/tachyon/pull/156#discussion_r3525357030

we should specify more structured digests to ensure input components of each digest are domain-separated.

zip-244 https://zips.z.cash/zip-0244 exhibits good patterns to follow

turbocrime:
work on this was done by #166 so i'm closing this (the intended actions occurred) even though spec is still being revised

PR #166, opened 2026-07-10
refactor: structured digests

turbocrime:
Work on #162.

Grows out of the digest and wire-format discussion on the bundle
transaction-format ZIP (#157) and the aggregation ZIP (#148).

Restructures the bundle and stamp digests into a tree modelled on ZIP 244. Each
named function is one protocol-defined BLAKE2b-256 hash with its own
personalization. Within a preimage at most one variable-length component is
hashed inline; every other input is fixed-length or a personalized sub-digest,
so no field boundary is ambiguous.

## Digest tree

`bundle_commitment` (`ZTxIdTachyonHash`, txid side):

```mermaid
graph TD
  vact@{ shape: procs, label: "bundle actions" }
  adgst@{ shape: trap-t, label: "action_descriptor_digest" }
  commit@{ shape: trap-t, label: "bundle_commitment" }
  vbal[value balance]

  vact --> adgst -->|hActionsTachyon| commit
  vbal -->|vBalanceTachyon| commit
```

`bundle_auth_digest` (`ZTxAuthTachyHash`, auth digest side):

```mermaid
graph TD
  state1{{0x01}}
  state2{{0x02}}

  vactsigs@{ shape: procs, label: "action sigs" }

  bsig[binding sig]

  vact@{ shape: procs, label: "bundle actions" }
  cvact@{ shape: procs, label: "adjunct actions" }
  sort((sort))
  adgst@{ shape: trap-t, label: "action_descriptor_digest" }

  proof[proof]
  sproof@{ shape: trap-t, label: "stamp_proof_digest" }
  anchor[anchor]
  tgs@{ shape: procs, label: "vTachygrams" }
  sdata@{ shape: trap-t, label: "stamp_data_digest" }

  cat((cat))
  wtxid[wtxid]

  state_impl{"impl"}
  stamp_impl{"impl"}

  auth@{ shape: trap-t, label: "bundle_auth_digest" }

  state1 ---|ProofStamp| state_impl 
  state2 ---|PointerStamp| state_impl 
  state_impl --> |tachyonBundleState| auth

  vactsigs -->|vActionSigsTachyon| auth
  bsig -->|bindingSigTachyon| auth

  vact & cvact --- sort --> adgst --- cat

  proof --> sproof
  sproof & anchor & tgs --> sdata --- cat

  cat ---|ProofStamp| stamp_impl
  wtxid ---|PointerStamp| stamp_impl

  stamp_impl -->|tachyonStampState| auth
```

## Ordering

Valid digests are over canonically ordered inputs. Flat-hashed lists only commit
unambiguously if their order is fixed.

- Digest utilities do not perform sorting.
- Bundle action sorting happens at construction only.
- Stamp tachygram sorting happens at construction only.
- Deserialization rejects non-canonical order.

### DIFF
--- a/crates/tachyon/src/action.rs
+++ b/crates/tachyon/src/action.rs
@@ -2,6 +2,7 @@
 
 use core::{cmp, marker::PhantomData};
 
+use alloc::vec::Vec;
 use corez::io::{self, Read, Write};
 use derive_more::{Debug, Display, Eq as TotalEq, PartialEq};
 use pasta_curves::{EpAffine, group::GroupEncoding as _};
@@ -40,18 +41,9 @@ impl Descriptor {
 
     /// Write an action descriptor in the consensus wire format.
     pub fn write<W: Write>(&self, mut writer: W) -> io::Result<()> {
-        serialization::write_ep_affine(&mut writer, &EpAffine::from(self.cv))?;
+        serialization::write_ep_affine(&mut writer, &self.cv.into())?;
         serialization::write_action_vk(&mut writer, &self.rk.0)?;
         Ok(())
-    }
-}
-
-impl From<Descriptor> for [u8; 64] {
-    fn from(desc: Descriptor) -> Self {
-        let mut bytes = [0u8; 64];
-        bytes[..32].copy_from_slice(&EpAffine::from(desc.cv).to_bytes());
-        bytes[32..].copy_from_slice(&<[u8; 32]>::from(desc.rk));
-        bytes
     }
 }
 
@@ -66,7 +58,11 @@ impl Ord for Descriptor {
         EpAffine::from(self.cv)
             .to_bytes()
             .cmp(&EpAffine::from(other.cv).to_bytes())
-            .then(<[u8; 32]>::from(self.rk).cmp(&<[u8; 32]>::from(other.rk)))
+            .then(
+                EpAffine::from(self.rk)
+                    .to_bytes()
+                    .cmp(&EpAffine::from(other.rk).to_bytes()),
+            )
     }
 }
 
@@ -205,7 +201,7 @@ impl Ord for Action {
     fn cmp(&self, other: &Self) -> cmp::Ordering {
         self.descriptor()
             .cmp(&other.descriptor())
-            .then_with(|| <[u8; 64]>::from(self.sig).cmp(&<[u8; 64]>::from(other.sig)))
+            .then_with(|| <[u8; 64]>::from(self.sig.0).cmp(&<[u8; 64]>::from(other.sig.0)))
     }
 }
 
@@ -237,14 +233,23 @@ impl Signature {
     }
 }
 
-impl From<[u8; 64]> for Signature {
-    fn from(bytes: [u8; 64]) -> Self {
-        Self(reddsa::Signature::<reddsa::ActionAuth>::from(bytes))
+impl FromIterator<Descriptor> for Vec<[u8; 64]> {
+    fn from_iter<I: IntoIterator<Item = Descriptor>>(iter: I) -> Self {
+        iter.into_iter()
+            .map(|desc| {
+                let mut desc_bytes = [0u8; 64];
+                desc_bytes[0..32].copy_from_slice(&EpAffine::from(desc.cv).to_bytes());
+                desc_bytes[32..64].copy_from_slice(&EpAffine::from(desc.rk).to_bytes());
+                desc_bytes
+            })
+            .collect()
     }
 }
 
-impl From<Signature> for [u8; 64] {
-    fn from(sig: Signature) -> [u8; 64] {
-        <[u8; 64]>::from(sig.0)
+impl FromIterator<Signature> for Vec<[u8; 64]> {
+    fn from_iter<I: IntoIterator<Item = Signature>>(iter: I) -> Self {
+        iter.into_iter()
+            .map(|sig| <[u8; 64]>::from(sig.0))
+            .collect()
     }
 }
--- a/crates/tachyon/src/action.rs
+++ b/crates/tachyon/src/action.rs
@@ -1,7 +1,8 @@
 //! Tachyon Action descriptions.
 
-use core::marker::PhantomData;
+use core::{cmp, marker::PhantomData};
 
+use corez::io::{self, Read, Write};
 use derive_more::{Debug, Display, Eq as TotalEq, PartialEq};
 
 use crate::{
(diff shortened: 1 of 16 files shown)


crates/tachyon/src/stamp/mod.rs
```diff
@@ -63,11 +71,62 @@ pub struct Stripped;
...
+    /// Write the stamp trailer in the consensus wire format.
+    fn write<W: Write>(&self, writer: W) -> io::Result<()>
+    where
+        Self: Sized;
+}
+
+impl StampState for PointerStamp {
+    fn stamp_digest(&self) -> [u8; 64] {
+        self.0
+    }
```
turbocrime, line 115: `PointerStamp::stamp_digest` returns wtxid directly.

i don't think there is any risk of wtxid collision with the outputs of `ProofStamp::stamp_digest` which is a concatenation of `action_descriptor_digest || stamp_data_digest` since all components involved are already domain-separated digests.

crates/tachyon/src/stamp/mod.rs
```diff
@@ -322,96 +471,90 @@ impl Stamp {
...
-            right.anchor,
+            right_anchor,
         ));
 
-        let tachygrams = [left.tachygrams, right.tachygrams].concat();
+        let merged_digests = [left_digests, right_digests].concat();
+        let tachygrams = [left_tachygrams, right_tachygrams].concat();
         let merged_tg_poly = TachygramSetPoly::from_iter(tachygrams.clone());
-        let merged_acts_poly = ActionSetPoly::from_iter([left_digests, right_digests].concat());
+        let merged_acts_poly = ActionSetPoly::from_iter(merged_digests.iter().copied());
```
LucidSamuel: clippy fails the `-D warnings` gate here (`from_iter_instead_of_collect`, exit 101) -> `merged_digests.iter().copied().collect::<ActionSetPoly>()`.

crates/tachyon/src/bundle/mod.rs
```diff
@@ -418,33 +434,34 @@ impl Plan {
...
     }
 
-    /// Apply externally-produced signatures (e.g. from FROST).
+    /// Apply externally-produced signatures.
     ///
     /// Validates each signature against the action's rk and the sighash.
-    /// Derives cv from each plan and produces the binding signature.
+    /// Derives cv from each plan and produces the binding signature. The
+    /// resulting actions are canonically sorted; each signature stays paired
+    /// with its descriptor, so sorting preserves the pairing.
```
LucidSamuel: code contradicts its own doc and `sign()` cause ("sorting preserves the pairing") is false for `apply_signatures`: it calls `descriptors()` which sorts descriptors *internally*, then zips against caller-ordered `sigs` so `sorted_desc[i]` pairs with `plan_order_sig[i]`, mismatched whenever canonical order ≠ insertion order.

crates/tachyon/src/bundle/mod.rs
```diff
@@ -473,162 +490,115 @@ impl Bundle<Unproven> {
...
+                )
+            })?;
+
+        let mut descriptors: Vec<action::Descriptor> = Vec::new();
+        for _ in 0..n_actions {
+            descriptors.push(action::Descriptor::read(&mut reader)?);
+        }
 
-        if head != BundleState::Stripped {
+        if !descriptors.is_sorted() {
```
LucidSamuel: non blocking: maybe a clean fix is to check `actions.is_sorted()` on Action Ord so the read invariant matches construction as this gate checks `descriptors.is_sorted()` (Descriptor Ord, sig-blind), but construction sorts on Action Ord (tiebreaks on the sig) ergo two actions with identical (cv, rk) and swapped sigs pass read, keep the same txid, but change auth_digest → a second valid wtxid.

crates/tachyon/src/action.rs
```diff
@@ -1,17 +1,73 @@
...
+    /// Read an action descriptor from the consensus wire format.
+    pub fn read<R: Read>(mut reader: R) -> io::Result<Self> {
+        let cv = value::Commitment(serialization::read_ep_affine(&mut reader)?);
+        let rk = public::ActionVerificationKey(serialization::read_action_vk(&mut reader)?);
+        Ok(Self { cv, rk })
+    }
+
+    /// Write an action descriptor in the consensus wire format.
+    pub fn write<W: Write>(&self, mut writer: W) -> io::Result<()> {
+        serialization::write_ep_affine(&mut writer, &self.cv.0)?;
```
LucidSamuel: inconsistent with the `From<Descriptor> for [u8;64]` right above it that uses `<[u8;32]>::from(...)`

LucidSamuel (changes requested):

crates/tachyon/src/stamp/mod.rs
LucidSamuel: agree and think the gap is that it's not structural cause a proof-stamp contrib (`hStampActionsTachyon || stamp_data_digest`) and a pointer wtxid share the same 64-byte slot with no committed tag saying which form it is, and `tachyonBundleState` (0x01/0x02) isn't absorbed into `bundle_auth_digest` at all. ZIP 244 commits the tx version/type for exactly this absorbing the state byte makes proof-vs-pointer part of the identity by construction rather than by hardness. cheap, closes it structurally.

crates/tachyon/src/stamp/mod.rs
turbocrime: `just lint` will use nightly clippy, and you should configure your IDE to also use nightly clippy. our `rust-toolchain.toml` configures stable, because we target the stable compiler, so you have to configure it manually.

`from_iter_instead_of_collect` was removed from clippy a month ago https://github.com/rust-lang/rust-clippy/commit/917d94a8ed5c542cf4c597bf2c48c0577fdec8fb

regardless, stable will accept it if i use `merged_digests.clone()` instead of `merge_digests.iter().copied()`, so you will see that lint disappear.

crates/tachyon/src/bundle/mod.rs
turbocrime: good catch

turbocrime: well. i think any case in which action signatures could slip by the descriptor check would necessarily be invalid.

for one, the actions would have to be duplicates, except for the signature. this would mean it's a double spend or double output. also, matching rk can't produce two different valid signatures.

so any such case would likely be caught by tachygram rules, signature validation, or other invariants.

still, it would be obviously wrong and rejectable at parse-time, so it's added.

turbocrime: i'm also removing the descriptor check as well, since it's totally satisfied by the later action check.

crates/tachyon/src/action.rs
turbocrime: i think i changed this before i saw your comment, and maybe the line number changed. not sure what you're referring to?

crates/tachyon/src/bundle/mod.rs
turbocrime: the caller must provide sigs in canonical order.

turbocrime: the docstring is now clarified

crates/tachyon/src/action.rs
LucidSamuel: now moot.

LucidSamuel (approved):
LGTM;

non blocking: leftover commented-out debug_assert! at bundle/mod.rs:168 -> delete or restore.

crates/tachyon/src/stamp/mod.rs
turbocrime: byte is added to the digest

crates/tachyon/src/bundle/mod.rs
turbocrime: added a duplicate check.

PR #167, opened 2026-07-11
fix: enforce `valueBalanceTachyon` rules, introduce `Value` types

### LINKED ISSUE #165: Enforce `valueBalanceTachyon` rules
The bundle ZIP draft (#104) adds two consensus rules on `valueBalanceTachyon`
that must be enforced:

- `valueBalanceTachyon` MUST be in the range between positive and negative
  `MAX_MONEY`, inclusive (bundle validity rule 7).
- A bundle with no actions MUST have `valueBalanceTachyon` equal to zero
  (bundle validity rule 8).

## Scope

- Implement checks at deserialization
- Implement checks at handling sites
- Action steps already enforce per-note value limits
- Merge steps are applicable in-bundle and cross-bundle, and may not be able to reason about this.
- Failing tests and edge-case tests

turbocrime:
Enforces the two `valueBalanceTachyon` consensus rules from the draft bundle ZIP that were previously unchecked:

- Range (rule 7): `valueBalanceTachyon` must fall within `-MAX_MONEY..=MAX_MONEY`.
- Zero-action (rule 8): a bundle with no actions must have a zero value balance.

Both rules are enforced by construction rather than rechecked at each call site. `Bundle::read` rejects an out-of-range or zero-action/nonzero-balance encoding at deserialization.

`bundle::ValueBalance` and `note::Value` are consolidated with the value-commitment trapdoor's value handling into one const-generic type, `value::Value<MIN, MAX>`, aliased `Balance`, `Positive`, and `Negative` for their respective ranges.

The value-commitment trapdoor and commitment types are renamed `value::Trapdoor` and `value::Commitment`; `Commitment` is now backed by `Ep` (projective) rather than `EpAffine` so it may derive `Add`, `Sub`, and `Sum`.

`NOTE_VALUE_MAX` is renamed `MAX_MONEY` to align with the spec name.

## Testing

- `just test`: full suite green, including boundary/overflow tests at the `Plan`, wire, and construction layers.
- `just lint`: clean.

Closes #165.

### DIFF
--- a/crates/tachyon/src/bundle/mod.rs
+++ b/crates/tachyon/src/bundle/mod.rs
@@ -64,7 +64,6 @@
 //! | `tachyonAggregateId`  | 64 bytes             | wtxid of the relevant aggregate          |
 
 use alloc::vec::Vec;
-use core::ops;
 
 use corez::io::{self, Read, Write};
 use derive_more::{Debug, Display, Eq as TotalEq, Error, From, PartialEq};
@@ -77,9 +76,9 @@ pub use crate::digest::blake2b::{AUTH_DIGEST_NO_BUNDLE, COMMIT_NO_BUNDLE};
 use crate::{
     ActionSetPoly,
     action::{self, Action},
+    constants::MAX_MONEY,
     digest::blake2b,
     keys::{private, public},
-    note,
     primitives::{ActionDigest, ActionDigestError, Anchor, effect},
     reddsa, serialization,
     stamp::{self, AggregateId, Stamp, Stripped, Unproven},
@@ -140,7 +139,7 @@ pub struct Bundle<S: StampState> {
     pub actions: Vec<Action>,
 
     /// Net value of spends minus outputs (plaintext integer).
-    pub value_balance: i64,
+    pub value_balance: ValueBalance,
 
     /// Binding signature over the transaction sighash.
     pub binding_sig: Signature,
@@ -207,7 +206,7 @@ pub enum CommitError {
     ActionDigest(#[error(not(source))] ActionDigestError),
     /// The value balance overflows the representable range.
     #[display("value balance overflow")]
-    BalanceOverflow(#[error(not(source))] BalanceError),
+    BalanceOverflow(#[error(not(source))] ValueBalanceOverflow),
 }
 
 /// Errors from bundle signature verification.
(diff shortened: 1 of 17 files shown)


crates/tachyon/src/bundle/mod.rs
```diff
@@ -756,12 +759,21 @@ impl<S: StampState> Bundle<S> {
...
 
     /// Verify the bundle's binding signature and all action signatures.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the bundle has no actions but a nonzero value balance.
     pub fn verify_signatures(&self, sighash: &[u8; 32]) -> Result<(), SignatureError> {
+        if self.actions.is_empty() && self.value_balance != ValueBalance::ZERO {
+            // Every construction path already guarantees this.
+            unreachable!("bundle with no actions cannot have nonzero value balance");
```
LucidSamuel: this `unreachable!` is reachable: `Bundle`'s fields are `pub`, so `Bundle { actions: vec![], value_balance: ValueBalance::try_from(1)?, .. }` is constructible and `zero_action_bundle_rejects_nonzero_balance` builds exactly that and asserts the panic, which falsifies the "every construction path already guarantees this" comment. and since it's the first statement, it aborts before any sig check, while `read_bundle_body` handles the identical predicate as a graceful `InvalidData`

crates/tachyon/src/bundle/mod.rs
```diff
@@ -781,56 +793,49 @@ impl<S: StampState> Bundle<S> {
...
-    type Output = Result<Self, BalanceError>;
-
-    fn add(self, rhs: note::Value) -> Self::Output {
-        self.0
-            .checked_add(u64::from(rhs).into())
-            .map(Self)
-            .ok_or(BalanceError)
+impl From<ValueBalance> for i64 {
+    fn from(balance: ValueBalance) -> Self {
+        assert!(
```
LucidSamuel: all three asserts just re-check what the newtype's `TryFrom` already guarantees at construction and since Value/ValueBalance are airtight (private inner, no arithmetic/serde), the outbound conversions can just be `balance.0`/`value.0` with no assert.

crates/tachyon/src/bundle/tests.rs
```diff
@@ -851,3 +854,197 @@ fn coverage_check_matches_stamped_action_set() {
...
+fn plan_value_balance_rejects_overflow_above_max_money() {
+    let rng = &mut StdRng::seed_from_u64(0);
+    let wallet = WalletSim::random(rng);
+    let ask = wallet.sk.derive_auth_private();
+    let spend_a = spend_plan_at(rng, &wallet, &ask, MAX_MONEY);
+    let spend_b = spend_plan_at(rng, &wallet, &ask, MAX_MONEY);
+    let bundle_plan = Plan::new(alloc::vec![spend_a, spend_b], alloc::vec![]);
+
+    assert_eq!(bundle_plan.value_balance(), Err(ValueBalanceOverflow));
+    assert!(matches!(
```
LucidSamuel: `assert!(matches!(result, Err(...)))` at tests.rs:912/917/935/940 (let-else + panic!), and it's inconsistent: other new tests in this PR use the correct form.

LucidSamuel (commented):
LGTM.
one thing to fix before merge: `verify_signatures` panics on constructible input (including your own #[should_panic] test).

crates/tachyon/src/bundle/mod.rs
turbocrime: it's only reachable if you manually assemble an invalid bundle. no normal code path reaches it, including deserialization.

i think it's fine to panic if state has been deliberately mutated in a way that violates specifications.

we should have an issue for auditing field visibility.

turbocrime: assertion removed, and 41fbeb7f0a8edd7fa33237a45bef6a7f6076154a demonstrates verify_signatures still rejects invalid value balances because a mutated field necessarily produces a signature that can't verify.

crates/tachyon/src/bundle/mod.rs
turbocrime: i considered this cheap defense-in-depth.

removed

LucidSamuel (approved):
GG;

Issue #164, opened 2026-07-08
Outputs could emit two tachygrams


turbocrime:
A spend emits two tachygrams (its present-epoch and next-epoch nullifiers); an
output emits one (its note commitment). The per-action arity is therefore
variable, and a stamp's tachygram count bears no fixed relation to the actions
it covers.

A variable relationship makes reconstruction from a candidate action set
ambiguous: a collector cannot know how many tachygrams a given action accounts
for, so a near-complete set is indistinguishable from the complete one short of
proof verification. This ambiguity motivated #149. With a fixed 2:1
relationship, we might remove `cActionsTachyon` and require total reconstruction
again.

With variable arity, a stamped bundle's public counts reveal its composition
(spends = nTachygrams − nActions). At fixed 2:1 the counts reveal nothing beyond
the action count, extending the existing role symmetry (cv hides an action's
sign; a tachygram does not distinguish a nullifier from a commitment).

## Scope

- **The second output tachygram.** Define what an output's second element is.
  It enters the pool accumulator and falls under the tachygram distinctness
  rules (per stamp, per block, and the two-epoch window), so it must be unique
  per output; a fixed dummy element collides immediately.
- **Circuit changes.** The output half of the action statement must produce and
  publish the second element; the proof attests it as it does the commitment.
- **Pool growth and sync cost.** Output-heavy usage roughly doubles its
  tachygram footprint: accumulator absorption, stamp size, and scan volume.
- **Spec changes across the ZIP drafts.** The shielded-protocol ZIP (#103) owns
  tachygram emission semantics; the bundle ZIP (#104) owns the stamp trailer
  (remove `cActionsTachyon`; the fail-fast reconstruction check of block
  validity folds into proof verification) and the digest inputs; the
  accumulator ZIP (#105) is structurally unchanged.

## Deliverables

- [ ] Design decision on the second output element
- [ ] `book/src/zips/tachyon-shielded-protocol.md`: emission semantics
- [ ] `book/src/zips/tachyon-bundle.md`: stamp trailer, block validity, digest
      inputs; state the fixed `nTachygrams = 2 * covered actions` relation
- [ ] `book/src/zips/zip-244.md`: digest input adjustments if the trailer changes
- [ ] Reference implementation in `crates/tachyon`

PR #168, opened 2026-07-11
refactor: outputs actions emit two tachygrams

turbocrime:
Work on #164

Stacks on #166, #167. Review after those merge.

The note commitment is now `(Fp, Fp)` produced by two squeezes of the Poseidon sponge.

All use locations now handle both elements.

An output's proof publishes both values as tachygrams.

### DIFF
--- a/crates/tachyon/src/action.rs
+++ b/crates/tachyon/src/action.rs
@@ -1,7 +1,8 @@
 //! Tachyon Action descriptions.
 
-use core::marker::PhantomData;
+use core::{cmp, marker::PhantomData};
 
+use corez::io::{self, Read, Write};
 use derive_more::{Debug, Display, Eq as TotalEq, PartialEq};
 
 use crate::{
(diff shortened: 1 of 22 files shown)


LucidSamuel (commented):
the crate's `Spec::CONSTANTS` doc says "a test re-derives the chain and cross-checks the table," but the existing `pinned_round_constants` tests only spot-check a few hand-copied `Fp::from_raw` literals, they don't independently re-derive the full tables. 

wrote the derivation test the doc describes: a `check_constants_derive` helper that independently re-derives every constant from the NUMS definition (`c0 = 0`; `ci = blake2b-512(LE index, personal = domain)` reduced via `FromUniformBytes<64>`) and asserts it equals the compiled `CONSTANTS`, trusting neither the committed `.bin` nor `tachyon_constants.py`. all three instances pass (~8288 constants), full suite green.

happy to fold it in here, or land it as a follow-up once this merges

turbocrime:
this landed in #182

PR #154, opened 2026-07-02
feat: mimc crate

turbocrime:
mimc crate for use by following PRs that may use it.

PR #171, opened 2026-07-15
refactor: flat mimc

turbocrime:
Experiment consuming #154

Should provide 128 nullifiers in a three-step derivation.

The master key becomes a tuple representing a round key and a whitening key.
Nullifiers are cipher outputs.

- `SboxStep` proves the round transition across interior columns
- `WrapStep` proves the row wrap plus whitening
- `NfDerive` binds the master key and converts to the typical header format


LucidSamuel:
this looks like it closes the old `nf_next parameter may need to come back` TODO from `main`: `nf_next` is no longer a free extra input to the stamp path, but comes through the bound derivation/spend flow. 

I think that is the right shape so just confirming this is intentional so we don’t carry the TODO forward @turbocrime

turbocrime:
i don't think we're going in this direction.

Issue #137, opened 2026-06-09
Bundle commitment is order-free over actions, so bundles are malleable to action reordering


turbocrime:
### Observation
The Tachyon bundle commitment (the `txid` contribution) commits to the action
**multiset**, not an ordered list. Reordering a bundle's actions therefore does
not change its `txid`, and because the action and binding signatures sign the
(order-free) sighash, the signatures stay valid under reordering. A third party
(e.g. a relayer) can permute `vActionsTachyon` / `vActionSigsTachyon` and produce
a valid bundle with the same `txid`.

### Mechanism
- `bundle_commitment` hashes `action_acc || value_balance`, where `action_acc` is
  the commitment to `∏(X − action_digest_i)` — permutation-invariant by
  construction.
- Each action signature and the binding signature sign the sighash, which derives
  from the order-free `txid`, so none of them bind action position.

### Effect
- Action ordering is malleable: same `txid`, different `wtxid`, same effect.
- This is `wtxid`-level (witness) malleability, not `txid` malleability — `txid`
  stays stable, which is the property tx chaining relies on.

### Why this may be intended
It mirrors how the tachygram list is treated: order is physical (committed in
`wtxid`), the multiset is effecting (committed in `txid`). Under that model,
action-reorder malleability is the expected analogue of witness malleability and
is tolerable. The adjunct→aggregate reference is unaffected: it is a `wtxid`
assigned by the miner at block assembly against the final top-level aggregate.

### To confirm / decide
- Confirm nothing relies on a canonical action order (mempool dedup, relay policy,
  any consumer that keys on action position).
- Decide whether to (a) document this as accepted `wtxid`-malleability, or
  (b) impose a canonical action ordering at serialization if any consumer needs a
  stable physical form.

### References
- `crates/tachyon/src/digest/blake2b.rs` — `bundle_commitment` (hashes
  `action_acc` coords + `value_balance`).
- `crates/tachyon/src/bundle/mod.rs` — `Bundle::commitment` / `Plan::commitment`
  (builds `ActionSetCommit` from the action digests); `verify_signatures` (action
  and binding sigs over the sighash).
- `book/src/transaction-identifiers.md` — `txid` = effecting data, `wtxid` =
  physical form.

PR #173, opened 2026-07-21
refactor: relax bundle ordering

turbocrime:
After #166, bundle deserialization required actions to be canonically sorted.
That check is redundant: order is already bound cryptographically. `Bundle::commitment`
hashes descriptors in wire order and feeds the sighash, so a signed bundle cannot
be reordered without breaking its signatures. This drops the sort requirement
(any order may be a valid, faithfully-committed encoding; `Plan` still emits
sorted bundles) and moves real double-spend prevention to where it belongs, the
stamp's proof.

## Wire and commitment

- `read` preserves action wire order and any duplicates. Multiplicity is
  significant to both the commitment and the proof, so it is carried faithfully
  rather than filtered at the parser.
- `Bundle::commitment` / `Bundle::descriptors` are documented order-sensitive and
  no longer sort.
- `ProofStamp` tachygram read stays strictly increasing (sorted and unique), the
  one canonical serialization of an unordered set.

## Verification

`Bundle<ProofStamp>::verify(rng, wtxid, adjuncts)` composes the full check and
returns `VerificationError`:

- `verify_pointers`: each adjunct's stamp digest equals the expected `wtxid`.
- `verify_coverage`: this bundle's and the adjuncts' descriptors are collectively
  unique (`DuplicateActions` otherwise), and `hStampActionsTachyon` matches that
  set. Duplicate detection now lives here, at verification, not at the parser.
- `verify_proof`: reconstructs the PCD header from public data and runs the Ragu
  verifier over the combined action multiset.

Coverage predicates are the cheap, proof-free pre-checks: `is_covering` (stamp
covers exactly the given descriptor multiset), `is_autonome` (covers exactly its
own actions), `is_aggregate` (does not). `covers` is renamed `is_covering` and
takes an `IntoIterator` multiset.

## Types

- `ProofStamp` coverage and tachygrams are `BTreeSet`, threaded as sets through
  the merge fold; `Plan::descriptors` returns a `BTreeSet` (the prover-side set),
  and `apply_signatures` takes a `BTreeMap<Descriptor, Signature>` matched against
  the descriptor set.
- Verification errors live on the bundle: `VerificationError`, `VerifyProofError`,
  `VerifyCoverageError`, `VerifyPointersError`. The stamp's own `verify` splits
  into crate-private `is_covering` + `verify_proof`.
- `Bundle` and `TachyonBundle` compare by `commitment` + `auth_digest`;
  `TachyonBundle` also orders by them. `as_dyn` borrows any stamped bundle as
  `&Bundle<dyn StampState>`.

## Tests

- `read_preserves_action_order`, `permuted_actions_change_commitment`: wire order
  survives round-trip and drives the commitment.
- `double_spend_obvious` / `double_spend_secret`: an obvious duplicate is caught
  at coverage; independently-randomized spends of one note pass the bundle layer,
  marking the bundle-vs-proof boundary.
- `double_spend_cannot_aggregate`, `duplicated_spend_cannot_inflate`,
  `cannot_forge_stamp_covering_duplicated_action`,
  `verify_cannot_distinguish_a_deduplicated_duplicate`: the proof, not the parser,
  rejects duplicated coverage.
- `autonome_verify_composes_all_checks`, `based_aggregate_with_two_adjuncts`,
  `verify_proof_disproves_uncovered_adjunct`,
  `verify_proof_rejects_action_shared_with_adjunct`: the composed verify flow over
  adjuncts.
- `auth_digest_invariants`: commitment held invariant while `auth_digest` moves.

### DIFF
--- a/crates/tachyon/src/bundle/mod.rs
+++ b/crates/tachyon/src/bundle/mod.rs
@@ -458,13 +458,13 @@ impl Bundle<ProofStamp> {
     /// Confirm `hStampActionsTachyon` represents this bundle's actions.
     #[must_use]
     pub fn is_autonome(&self) -> bool {
-        self.covers(&[])
+        self.stamp.covers(&self.descriptors())
     }
 
     /// Confirm `hStampActionsTachyon` does not represent this bundle's actions.
     #[must_use]
     pub fn is_aggregate(&self) -> bool {
-        !self.covers(&[])
+        !self.stamp.covers(&self.descriptors())
     }
 }
 
--- a/crates/tachyon/src/bundle/mod.rs
+++ b/crates/tachyon/src/bundle/mod.rs
@@ -143,15 +143,15 @@ impl<T: sealed::Sealed> BundleState for T {}
 /// A Tachyon transaction bundle parameterized by bundle state `S`.
 #[derive(Clone, Debug, PartialEq, TotalEq)]
 pub struct Bundle<S: BundleState + ?Sized> {
+    /// Net value of spends minus outputs (plaintext integer).
+    pub value_balance: value::Balance,
+
     /// Actions (cv, rk, sig).
     ///
     /// The bundle commitment is sensitive to the order of actions. The [`Plan`]
     /// utility in this crate sorts actions by descriptor, but other
     /// implementations may create any arbitrary ordering.
     pub actions: Vec<Action>,
-
-    /// Net value of spends minus outputs (plaintext integer).
-    pub value_balance: value::Balance,
 
     /// Binding signature over the transaction sighash.
     pub binding_sig: Signature,
@@ -481,7 +481,7 @@ impl Bundle<ProofStamp> {
     #[must_use]
     pub fn is_aggregate(&self) -> bool {
         let desc_bytes = self.descriptors().into_iter().collect::<Vec<[u8; 64]>>();
-        blake2b::action_descriptor_digest(&desc_bytes) == self.stamp.actions
+        blake2b::action_descriptor_digest(&desc_bytes) == self.stamp.coverage
     }
 }
 
@@ -553,8 +553,8 @@ impl<S: StampState> Bundle<S> {
         let stamp = S::read(&mut reader)?;
 
         Ok(Self {
-            actions,
             value_balance,
+            actions,
             binding_sig,
             stamp,
         })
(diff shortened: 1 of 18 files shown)


crates/tachyon/src/bundle/mod.rs
```diff
@@ -535,19 +500,27 @@ impl<S: StampState> Bundle<S> {
...
-            descriptors.push(action::Descriptor::read(&mut reader)?);
-        }
+            let descriptor = action::Descriptor::read(&mut reader)?;
 
-        let mut unique_descriptors = BTreeSet::new();
-        for descriptor in &descriptors {
-            if !unique_descriptors.insert(descriptor) {
+            // TODO: do we care about deduplication here?
+            // we could improve or delete this check.
+            if descriptors.contains(&descriptor) {
```
LucidSamuel: think we can either restore it as a BTreeSet (cheap, O(n log n)) or drop the check entirely, but not the
linear scan, i'd keep the BTreeSet.

crates/tachyon/src/stamp/mod.rs
```diff
@@ -180,38 +181,60 @@ impl StampState for ProofStamp {
...
+        // TODO: assert a reasonable maximum, to allow pre-allocation?
+        let n_tachygrams = usize::try_from(serialization::read_compactsize(&mut reader)?)
+            .map_err(|err| io::Error::new(io::ErrorKind::InvalidData, err))?;
+
+        let mut tachygrams: Vec<Tachygram> = Vec::new();
+        for _ in 0..n_tachygrams {
+            let tg = Tachygram::from(serialization::read_fp(&mut reader)?);
+
+            match tachygrams.last().map(|last| last.cmp(&tg)) {
+                None | Some(Ordering::Less) => tachygrams.push(tg),
```
LucidSamuel: non blocking: inlining the tachygram read leaves `read_fp_list` with no callers (only `write_fp_list` remains). maybe a note it's kept for write symetry or drop.

LucidSamuel (approved):
looks good #137 worth linking/closing it here.

one thing i'd change (the restored dedup check is O(n²)) and a couple of non-blocking notes below.

crates/tachyon/src/bundle/mod.rs
turbocrime: order of descriptors is not constrained here, so i don't want to use a BTreeSet. i think i'll simply remove the check, and make sure tests demonstrate that rejection of a bundle like this is enforced in other ways.

crates/tachyon/src/stamp/mod.rs
turbocrime: removed both.

Issue #123, opened 2026-05-26
Audit all handling of curve points and field elements to eliminate ambiguity


turbocrime:
as we near some state resembling 'completion' we should thoroughly review all use of `Fp`/`Fq`/`EpAffine`/`EqAffine` and all local or imported newtypes wrapping field elements or points.

- [ ] compile list of any raw value handling
- [ ] compile list of local newtypes
- [ ] compile list of imported newtypes (may be inobvious)
- [ ] identify imported impls, serde, binary reprs
- [ ] identify local impls, serde, binary reprs

pay special attention to

- digests
- keys
- polynomials
- commitments
- bundle serialization
- header encoding

most raw value handling should be replaced with a newtype. most newtype inners should be private.

ensure unambiguous binary repr in all contexts. justify any compressed representation or bit decomposition.

confirm correctness of all logic depending on signed (+/-) values, and justify any logic that is sign-ambiguous.

justify choice of field/scalar and element/point/affine.

include all justification and reasoning either in docstrings or in book.

PR #174, opened 2026-07-22
fix: mask sign bit when extracting ak x-coordinate

LucidSamuel:
`PaymentKey::derive` reinterpreted the compressed RedPallas verification key encoding (canonical x plus the y-sign bit) as an `Fp`, which panics in `Fp::from_repr` whenever the sign bit is set. 

mask byte 31 bit 7 to recover the canonical x-coordinate, preserving the identity encoding as `Fp::ZERO`
and closing the `CtOption::expect` lint hole. The honest key path already sign-normalizes `ak`, so this is a latent panic surface, not an active bug.

also adds regression tests for a sign-bit-set key and the identity key. part of #123.

### DIFF
--- a/crates/tachyon/src/keys/note.rs
+++ b/crates/tachyon/src/keys/note.rs
@@ -1,7 +1,7 @@
 //! Note-related keys: NullifierKey, PaymentKey.
 
 use derive_more::Debug;
-use ff::PrimeField as _;
+use ff::{Field as _, PrimeField as _};
 use pasta_curves::Fp;
 
 use super::{ggm::NoteMasterKey, proof::SpendValidatingKey};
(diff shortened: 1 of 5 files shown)


crates/tachyon/src/keys/note.rs
```diff
@@ -85,19 +85,91 @@ impl PaymentKey {
...
+        // It must derive from the masked x-coordinate, independently recomputed.
+        let mut masked = vk_bytes;
+        masked[31] &= 0b0111_1111;
+        let expected_x = Option::from(Fp::from_repr(masked)).expect("masked x is canonical");
+        assert_eq!(pk.0, poseidon::payment_key(expected_x, nk.0));
+    }
+
+    /// The identity key encoding `[0; 32]` must map to `Fp::ZERO` (not error) —
+    /// the behaviour the sign-bit mask deliberately preserves. reddsa permits
+    /// identity verification keys.
```
turbocrime: reddsa permits identity keys, but we do not need and probably should not support identity ak.

crates/tachyon/src/keys/note.rs
```diff
@@ -85,19 +85,91 @@ impl PaymentKey {
...
+/// Feeding the raw compressed bytes to [`Fp::from_repr`] directly instead
+/// panics whenever the y-sign bit is set: those bytes encode `x + 2^255 > p`,
+/// which is not a canonical field element. The honest `SpendingKey` path
+/// sign-normalizes `ak` (§5.4.7.1) so the bit is always clear, but an
+/// un-normalized valid key — e.g. one parsed from the wire — would trip it.
+fn vk_x_coordinate(ak: &SpendValidatingKey) -> Fp {
+    let mut bytes: [u8; 32] = ak.0.into();
+    // Clear the y-sign bit (byte 31, bit 7), leaving the canonical x-coordinate.
+    bytes[31] &= 0b0111_1111;
+    Option::from(Fp::from_repr(bytes)).unwrap_or(Fp::ZERO)
```
turbocrime: if this zero fallback is unreachable as the doc claims, we should panic instead of silently succeeding with an incorrect result.

crates/tachyon/src/keys/note.rs
```diff
@@ -85,19 +85,91 @@ impl PaymentKey {
...
+/// low 255 bits plus the y-sign in byte 31, bit 7 (Zcash protocol spec
+/// §5.4.9.7). Clearing that sign bit leaves the canonical x, which is always a
+/// valid `Fp`, so the conversion cannot fail — the fallback is unreachable. The
+/// identity encoding `[0; 32]` maps to `Fp::ZERO`.
+///
+/// Feeding the raw compressed bytes to [`Fp::from_repr`] directly instead
+/// panics whenever the y-sign bit is set: those bytes encode `x + 2^255 > p`,
+/// which is not a canonical field element. The honest `SpendingKey` path
+/// sign-normalizes `ak` (§5.4.7.1) so the bit is always clear, but an
+/// un-normalized valid key — e.g. one parsed from the wire — would trip it.
```
turbocrime: presently, i don't think we will actually wind up transmitting `ak` on the wire. `ProofAuthorizingKey` is useless once we discard GGM.

regardless: any noncanonical `ak` should be *refused* instead of silently transformed into *a different key*.

turbocrime (changes requested):
a `SpendValidatingKey` only comes from `SpendAuthorizingKey::derive_auth_public`

a `SpendAuthorizingKey` only comes from `SpendingKey::derive_auth_private`, which normalizes sign

https://github.com/tachyon-zcash/tachyon/blob/a0abbdaae72648aa260274c5ad5acc1a1c6218e8/crates/tachyon/src/keys/private.rs#L91-L99

there will be no uncontrolled input, since this is private key material,
produced by a derivation we control, from more private key material we already
hold.

so the bit is always clear.

this is an invariant that should panic on violation.

turbocrime:
pushed c4035b06158829d3dd034fd8f38fd14d9b7f931e to deny `CtOption::expect` and `CtOption::unwrap`

LucidSamuel:
Superseded by #180

PR #175, opened 2026-07-23
test: guard digest domain seperator reuse

LucidSamuel:
adds executable distinctness checks for the current digest contexts
-a registry of every Poseidon (tag, arity) and BLAKE2b (personal, domain byte) hash context, with a pairwise-separation test in each digest module. BLAKE2b tags are compared as the zero-padded 16-byte block the hash actually keys on.

also documents the known `nf_master/nf_prefix` GGM reuse (shared `Tachyon-NfPrefix` at arity 3) as an explicit allowlist entry rather than a failure. #171 removes that path, and should update the registry when it lands (drop the allowlist, add the MiMC domains).

~ this is a guard ahead of the full registry/spec work.

### DIFF
--- a/crates/tachyon/src/digest/blake2b.rs
+++ b/crates/tachyon/src/digest/blake2b.rs
@@ -236,6 +236,114 @@ pub(crate) fn bundle_auth_digest(
     })
 }
 
+#[cfg(test)]
+mod domain_separation {
+    //! Registry of every BLAKE2b hash context and a check that they are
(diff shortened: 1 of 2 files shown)


turbocrime:
this feels incomplete and i'm not sure it's a good solution. also, it's a moving target.

let's defer this kind of thing until the digests are completely settled.

LucidSamuel:
closing to park this per the deferral

Issue #52, opened 2026-03-30
Griefing vectors


TalDerei:
Fee increase mechanism to disincentivize griefing vectors (eg. unattractive transactions forcing validators to hold extra nullifier history).

PR #176, opened 2026-07-23
fix: fail fast on stamp merge coverage mismatch

LucidSamuel:
avoids wasted merge proving work when an aggregator is given descriptor lists that do not match the input stamps: `ProofStamp::merge` now compares each side's descriptors against the stamp's covered-actions digest and fails fast with `CoverageMismatch` (carrying both digests) before proving. 

resolves the in-code TODO; the proof system still enforces the same property, this just moves the rejection before the expensive work.

~refs #52 and also touches `stamp/mod.rs` so anticipates a small rebase if #173 lands first

### DIFF
--- a/crates/tachyon/src/stamp/mod.rs
+++ b/crates/tachyon/src/stamp/mod.rs
@@ -607,13 +607,24 @@ impl ProofStamp {
     /// The action digests for the merge proof and the merged
     /// `covered_actions` are both derived from the descriptor lists.
     ///
-    /// TODO: confirm desc list against stamp? it's forbidden by the proof
-    /// system, but we might want to fail early.
+    /// In debug builds, asserts that each descriptor list matches its stamp's
+    /// covered-actions digest. The proof system rejects a mismatched merge
+    /// regardless, but only after the merge proving work has been spent; a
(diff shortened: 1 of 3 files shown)


turbocrime:
i think this is not necessary. an aggregator can be expected to deliberately select suitable inputs. by the time they call `merge`, they know what they are doing, and we don't need to recompute the check.

this might be suitable as a debug_assert.

LucidSamuel:
noted, reworked to the debug_assert form (rebased onto #173 uses `is_covering` and the BTreeSet signature, resolves the TODO), if even this feels like noise, equally happy to close cos the proof system's rejection is the real gate either way.

LucidSamuel:
@turbocrime rebased onto main (post-#188) with the same `debug_assert` form you suggested, resolves the merge TODO. still equally happy to close if it reads as noise.

turbocrime:
i think i'd actually prefer to omit the debug_assert and let the proof fail.

this is really a test gap: we've failed to represent the scenario in which a caller did not respect the caller contract.

we have other tests of this method resulting in proof failures, and maybe we should add one for this specific case.

issued as https://github.com/tachyon-zcash/tachyon/issues/208

PR #177, opened 2026-07-23
fix: return error for identity commitment in anchor absorption

LucidSamuel:
`Anchor::next_stamp` panicked on an identity stamp commitment via `expect` on a `CtOption`, evading the panic lints. converted to `Result<Anchor, AnchorError>` so the identity point has no affine coordinates to absorb. Honest `TachygramSetPoly` commitments generated by this crate should also not be identity as the committed set polynomial is monic, so an identity commitment would correspond to a nontrivial linear relation among the fixed commitment generators, making this a defensive error surface rather than a reachable state, but a public API on the consensus anchor path shouldn't carry a latent panic.

step witnesses (`AnchorSeed`, `UnspentSeed`, `SpendableInit`) map the error to `InvalidWitness` plus a rejection regression test.

Part of #123


turbocrime (commented):
a `TachygramSetCommit` that reaches consensus (and thus `Anchor::next_stamp`) can never be empty. i expect it will change to an `EpAffine` inner which will express this better.

i do think `Anchor` needs improvements but i'm not sure this is it.

also, there is possibly a significant `Anchor` refactor incoming, so this may be obviated soon.

LucidSamuel:
makes sense, closing; will add a reminder to revisit after the anchor refactor lands.

PR #183, opened 2026-08-11
refactor: indexed multiset

turbocrime:
nullifier sequences are a multiset of `(nf, epoch)` encodings.

- `NfDerive` emits a window of 16 consecutive nullifiers
- `UnspentBind` confirms a subsequence by quotient

### DIFF
--- a/crates/tachyon/src/collections/indexed_multiset.rs
+++ b/crates/tachyon/src/collections/indexed_multiset.rs
@@ -1,21 +1,20 @@
 //! # Indexed multiset
 //!
-//! An indexed multiset may contain any number of unique or repeated nonzero
-//! members at each nonzero index. It is a multiset of `(index, member)`
-//! tuples.
+//! An indexed multiset may contain any number of unique or repeated members at
+//! each index. Conceptually, it is a multiset of `(index, member)` tuples.
 //!
 //! ## Irreducible encoding
 //!
-//! Construct a single-member indexed multiset of member $m$ at index
-//! $i$ as the polynomial
+//! Construct a single-member indexed multiset of member $m$ at index $i$ as the
+//! polynomial
 //!
 //! $$
-//!   F_{i,m}(X) = (iX + m)^3 - c
+//!   F_{i,m}(X) = (m + (i+1)X)^3 - c
 //! $$
 //!
 //! where $c = 2$ is selected because it is the smallest cubic non-residue.
 //!
-//! Since $(iX + m)^3 \neq c$ the encoding is irreducible.
+//! Since $(m + (i+1)X)^3 \neq c$ the encoding is irreducible.
 //!
 //! ## Product composition
 //!
@@ -60,14 +59,14 @@
 //! Precisely, if
 //!
 //! $$
-//!   \frac{i_1X + m_1}{i_2X + m_2} \in \{1, \zeta, \zeta^2\}
+//!   \frac{m_1 + (i_1+1)X}{m_2 + (i_2+1)X} \in \{1, \zeta, \zeta^2\}
 //! $$
 //!
 //! then $(i_1, m_1)$ cannot be distinguished from $(i_2, m_2)$ in an
 //! indexed multiset.
 //!
-//! Specifying $i_\mathsf{max}$ below $\lfloor\sqrt{p/3}\rfloor$ is sufficient
-//! to prevent collisions.
+//! Specifying $i_\mathsf{max} + 1$ below $\lfloor\sqrt{p/3}\rfloor$ is
+//! sufficient to prevent collisions.
 
 #![allow(clippy::min_ident_chars, reason = "just for fun")]
 
--- a/crates/tachyon/src/collections/indexed_multiset.rs
+++ b/crates/tachyon/src/collections/indexed_multiset.rs
@@ -73,8 +73,6 @@
 
 extern crate alloc;
 
-use core::{iter, num::NonZero};
-
 use ff::Field as _;
 use pasta_curves::Fp;
 use ragu::Polynomial;
@@ -85,7 +83,8 @@ use super::poly_mul;
 const NON_RESIDUE: Fp = fp!(0x02);
 
 #[must_use]
-fn encode_single(i: Fp, m: Fp) -> Polynomial {
+fn encode_single(idx: u64, m: Fp) -> Polynomial {
+    let i = Fp::from(idx) + Fp::ONE;
     // writing out expanded coefficients for $F(X) = (iX + m)^3 - c$ is
     // cheaper than constructing a linear $f(X) = iX + m$ and then cubing it.
     Polynomial::from_coeffs(
@@ -100,39 +99,29 @@ fn encode_single(i: Fp, m: Fp) -> Polynomial {
 }
 
 #[must_use]
-fn direct_eval_single(i: Fp, m: Fp, x: Fp) -> Fp {
+fn direct_eval_single(idx: u64, m: Fp, x: Fp) -> Fp {
+    let i = Fp::from(idx) + Fp::ONE;
     ((i * x) + m).pow([3]) - NON_RESIDUE
 }
 
-/// Encode the provided members consecutively.
-pub(crate) fn encode(start_idx: NonZero<u64>, members: impl IntoIterator<Item = Fp>) -> Polynomial {
-    let i_m = iter::successors(Some(start_idx), |idx| idx.checked_add(1))
-        .map(|idx| Fp::from(idx.get()))
-        .zip(members);
-
-    i_m.fold(
+/// Encode the provided indexed members.
+pub(crate) fn encode(members: impl IntoIterator<Item = (u64, Fp)>) -> Polynomial {
+    members.into_iter().fold(
         Polynomial::from_coeffs([Fp::ONE].to_vec()),
-        |acc, (i, m)| poly_mul(&acc, &encode_single(i, m)),
+        |acc, (idx, m)| poly_mul(&acc, &encode_single(idx, m)),
     )
 }
 
 /// Evaluate the indexed multiset without building the polynomial.
-pub(crate) fn direct_eval(
-    start_idx: NonZero<u64>,
-    members: impl IntoIterator<Item = Fp>,
-    x: Fp,
-) -> Fp {
-    let i_m = iter::successors(Some(start_idx), |idx| idx.checked_add(1))
-        .map(|idx| Fp::from(idx.get()))
-        .zip(members);
-
-    i_m.fold(Fp::ONE, |acc, (i, m)| acc * direct_eval_single(i, m, x))
+pub(crate) fn direct_eval(members: impl IntoIterator<Item = (u64, Fp)>, x: Fp) -> Fp {
+    members
+        .into_iter()
+        .fold(Fp::ONE, |acc, (idx, m)| acc * direct_eval_single(idx, m, x))
 }
 
 #[cfg(test)]
 mod tests {
     use alloc::vec::Vec;
-    use core::iter;
 
     use rand::{RngExt as _, SeedableRng as _, rngs::StdRng};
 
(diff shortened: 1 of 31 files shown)


crates/tachyon/src/primitives/epoch.rs
```diff
@@ -42,12 +46,52 @@ impl EpochIndex {
...
+#[expect(
+    clippy::as_conversions,
+    clippy::cast_possible_truncation,
+    reason = "the group and window widths are small constants, and flooring to \
+              the last whole group is the intended bound"
+)]
+impl EpochGroup {
+    /// The group's first epoch.
+    #[must_use]
+    pub const fn start_epoch(self) -> EpochIndex {
```
LucidSamuel: checking if I missed something, but `EpochGroup` isn't injectively pinned by the emitted base in release builds. `start_epoch()` does a wrapping `u32` multiply, so `EpochGroup(0)` and `EpochGroup(1<<30)` both announce epoch 0 while `nullifier_group` absorbs their distinct raw indices (`Fp::from(group.0)`), forking the derived window which falsifies the `NfDerive` soundness note at delegation.rs:147 ("base = NF_DERIVATION_GROUP · w₀ is injective, so the emitted epoch determines w₀"), and since `EpochGroup(pub u32)` is publicly constructible with no `EPOCH_MAX` bound in `NfDerive`, a prover picks the aliased group freely.

maybe we can enforce the bound inside `NfDerive` (witness-helper checks are bypassable via the public constructor): reject any group whose full window exceeds `EPOCH_MAX`, use checked/widening arithmetic for the base, offsets, and exclusive end, and add a release regression covering the alias. i have a ready-made one (verifies release/panics debug) i can drop in or as a follow-up PR if need be.

LucidSamuel: maybe i'm missing something, but i don't think the epoch group is actually pinned by the epoch it announces in release builds.

`start_epoch()` multiplies the group index by 4 as a plain u32, which can overflow so `EpochGroup(0)` and `EpochGroup(1<<30)` both announce epoch 0, but the sponge hashes the raw group index, so they produce different nullifier windows that breaks the `NfDerive` soundness comment (delegation.rs:147) that says the emitted epoch determines the group and `EpochGroup` is a public tuple with no `EPOCH_MAX` bound, so a prover can pick the aliased group directly.

fix probably needs to live in `NfDerive` (the public constructor bypasses the witness helpers): reject groups whose window exceeds E`POCH_MAX` and use checked math for the base/offsets/end. i've got a regression test (passes release, panics debug), happy to PR it against this branch if needed.

LucidSamuel (approved):
LGTM and based!

small nit: release-only soundness break (executed as a double-spend).

crates/tachyon/src/digest/poseidon.rs
```diff
@@ -79,34 +83,41 @@ pub fn pad_tachygram(rcm: Fp, pk: Fp, value: u64, psi: Fp) -> Fp {
...
-pub fn nf_prefix(prefix_prev: Fp, step: u8) -> Fp {
-    hash::<3>([
-        Fp::from_u128(u128::from_le_bytes(*NULLIFIER_PREFIX_DOMAIN)),
-        prefix_prev,
-        Fp::from(u64::from(step)), // TODO: chunk some booleans by arity?
-    ])
-}
-
+/// Nullifiers derived per sponge.
+pub const NF_DERIVATION_GROUP: usize = 4;
```
alxiong: nit: should at least ends with `_SIZE`, maybe `NF_GROUP_SIZE` or `NF_DERIVATION_BATCH_SIZE` etc.

crates/tachyon/src/primitives/epoch.rs
```diff
@@ -22,6 +22,10 @@ pub struct EpochIndex(pub u32);
 #[into(u64)]
 pub struct EpochDiff(u32);
 
+/// The index of a group of [`NF_DERIVATION_GROUP`] consecutive epochs.
+#[derive(Clone, Copy, Debug, From, Into, Ord, PartialEq, PartialOrd, TotalEq)]
+pub struct EpochGroup(pub u32);
+
```
alxiong: personal take: i don't find this abstraction necessary or future-proof. 

if we are always going to use simple fixed-size batching/chunking, then a simple division is fine. forcing to use `fn epoch_group()` doesn't feel ergonomic.

a more future-proof api design would be introduce an iterator constructed with a specified `Range`, and keeping state of current epoch, lazily batch-derive internally.

Can we go with `EpochIter` instead?

crates/tachyon/src/collections/multiseq.rs
```diff
@@ -0,0 +1,196 @@
+//! # Multisequence
+//!
+//! A multisequence may contain any number of unique or repeated nonzero members
+//! at each nonzero index. Conceptually, it is a multiset of `(index, member)`
+//! tuples.
```
alxiong: we would rename this module as `indexed_multiset`, because it's not guaranteed to be a sequence even. and "indexed multiset" doesn't have an extra "mental translation/indirection" regarding its definition.

"sequence" has the conventional connotation that it's a consecutive indexed series of values, which is not what we have. we don't guaranteed consecutive indexes.

crates/tachyon/src/collections/indexed_multiset.rs
```diff
@@ -0,0 +1,196 @@
...
+
+use super::poly_mul;
+
+const TWO: [u64; 4] = [2, 0, 0, 0];
+const THREE: [u64; 4] = [3, 0, 0, 0];
+
+#[must_use]
+fn encode_single(i: Fp, m: Fp) -> Polynomial {
+    // writing out expanded coefficients for $F(X) = (iX + m)^3 - \zeta$ is
+    // cheaper than constructing a linear $f(X) = iX + m$ and then cubing it.
```
alxiong: incorrect, it should be `(i+1) X + m) ^ 3 - c`, we should accept arbitrary index `i: Fp` but we shouldn't let leading coefficient to be zero. 

i.e. we should NOT to expect caller to pass in `i+1`. but `+1` is internal details.

crates/tachyon/src/collections/multiseq.rs
```diff
@@ -0,0 +1,196 @@
...
+
+use core::{iter, num::NonZero};
+
+use ff::Field as _;
+use pasta_curves::Fp;
+use ragu::Polynomial;
+
+use super::poly_mul;
+
+const TWO: [u64; 4] = [2, 0, 0, 0];
```
alxiong: 1. rename this to either `C` or `SMALL_NQR` 
2. add unit test that this value is NQR.

crates/tachyon/src/collections/indexed_multiset.rs
```diff
@@ -0,0 +1,196 @@
...
+    )
+}
+
+#[must_use]
+fn direct_eval_single(i: Fp, m: Fp, x: Fp) -> Fp {
+    ((i * x) + m).pow(THREE) - Fp::from_raw(TWO)
+}
+
+/// Encode the provided members consecutively.
+pub(crate) fn encode(start_idx: NonZero<u64>, members: impl IntoIterator<Item = Fp>) -> Polynomial {
```
alxiong: since i internalize this primitive as `indexed_multiset`, thus the constructor api i have in mind is:

`fn encode(members: impl IntoIterator<Item = (u64, Fp)>) -> Polynomial`

crates/tachyon/src/collections/multiset.rs
```diff
@@ -0,0 +1,12 @@
+extern crate alloc;
+
+use alloc::vec::Vec;
+
+use pasta_curves::Fp;
+use ragu::{Polynomial, poly_with_roots};
+
+/// Encode the provided members as roots.
+pub(crate) fn encode(members: impl IntoIterator<Item = Fp>) -> Polynomial {
```
alxiong, line 9: remove this? don't think this special case deserve a dedicated constructor.

crates/tachyon/src/collections/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+pub(crate) mod multiseq;
+pub(crate) mod multiset;
+
+fn trim(coeffs: &mut Vec<Fp>) {
+    if let Some(last_nonzero_idx) = coeffs.iter().rposition(|co| co != &Fp::ZERO) {
+        coeffs.truncate(last_nonzero_idx + 1);
+    }
+}
+
+pub(super) fn poly_mul(input_a: &Polynomial, input_b: &Polynomial) -> Polynomial {
```
alxiong, line 18: i'm not sure why these helper func is necessary.

alxiong (commented):
I did a first-pass, mostly on the indexed multiset, but not yet the circuit/stamp part (which i will get to in a second pass).

For the encoded polynomial for an indexed multiset, I believe the right approach is to delay "realization into `ragu::Polynomial` as late as `.commit()` time. We should instead just keep the `HashSet<(usize, Fp)>` as long as we could because majority of the operation are much cheaper by operating on this. 

The most important reason is to avoid polynomial division when computing the quotient witness, if we keep it as a HashSet, then the `q(X)` computation is as easy as set subtraction, not even field ops involved. 

multiplication is also cheap set union. evaluation is what you already have as `direct_eval`, but i'd just make it a method as `IndexedMultisetPoly::eval()` which computes in streaming fashion. this is definitely faster than `.eval()` on a realized `ragu::Poly` because we can massively parallelize the evaluation.

alxiong:
another comment is: can we split up this PR into 2 instead? we can quickly merge the `Poseidon` based nullifier first, then the "indexed multiset" relation update is in a separate PR? 
(maybe I didn't take into account of the stamp generation logic that demands update in lock-step, so this is a suggestion, not a request)

crates/tachyon/src/primitives/epoch.rs
turbocrime: okay, i removed the group newtype, now i just enforce_zero(e%4)

there is only so far the mock can take this. ultimately the real circuit will need some kind of bit-decomp

turbocrime: dropping the group newtype and switching to e%4=0 now demonstrates the intent properly. but again constraining a maximum will require some kind of bit-decomp and i think it's fine to just leave that out of the mock for now

crates/tachyon/src/digest/poseidon.rs
turbocrime: deleting this const in favor of importing ragu's `PoseidonFp::RATE`

crates/tachyon/src/primitives/epoch.rs
turbocrime: deleted the newtype EpochGroup

crates/tachyon/src/collections/multiseq.rs
turbocrime: constructively, it will always be constructed as a contiguous sequence in our implementation, and then later it's possible to produce a union of "multiple" disjoint/overlapping sequences

this is what the name was intended to convey, even though it's technically capable of more.

i will transition the terminology

crates/tachyon/src/collections/indexed_multiset.rs
turbocrime: this is a private utility function and the public api enforces Nonzero parameter

the nonzero requirement is an implementation detail of the scheme and i'm choosing to abstract it at this location

crates/tachyon/src/collections/multiset.rs
turbocrime: this is just a relocation of existing multiset stuff to mirror the collections module created to colocate new sequence utils. this is not a new thing

crates/tachyon/src/collections/mod.rs
turbocrime: it's not

turbocrime:
> another comment is: can we split up this PR into 2 instead? we can quickly merge the `Poseidon` based nullifier first, then the "indexed multiset" relation update is in a separate PR?

i made this one PR because the introduction of window-aligned derivation encouraged the adoption of a sequenc eprimitive that allowed subsequence selection

i'll split it out

turbocrime:
> We should instead just keep the `HashSet<(usize, Fp)>` as long as we could because majority of the operation are much cheaper by operating on this.

a bare set can't represent multiplicity and hashset is not available in no-std

code that wants to handle a simple set type can use a BTreeSet (or std consumers can use a HashSet) and transition into the polynomial format using the provided utilities when they want to

crates/tachyon/src/collections/multiseq.rs
turbocrime: renamed to NON_RESIDUE locally already

crates/tachyon/src/collections/indexed_multiset.rs
turbocrime: renamed to indexed_multiset

turbocrime: aligning on the tuple iter signature

crates/tachyon/src/collections/mod.rs
turbocrime: but the imported poly tools do pad with zeros and don't bother to trim, and that's annoying, and if i don't write the helpers i have to write trim everywhere. can fix later.

alxiong: fixed in #201

PR #185, opened 2026-08-13
refactor: anchor commits to epoch

turbocrime:
the anchor now digests the epoch index at every increment, and no longer changes
when passing an empty block.

the anchor's domain-separated epoch tick remains present.

- removed `EmptyBlockSeed`
- removed `EmptyBlockUnspentSeed`
- removed `UnspentEpochFuse`
- `EndEpochUnspentSeed` is added to transit an epoch boundary
- `SpendableHeader` now carries an epoch field
- `SpendableInit` no longer requires an `AnchorChain` input to bind its epoch

some other renames:
- renamed header `Unspent` to `ArbitraryUnspent`
- renamed header `VerifiedUnspent` to `Unspent`
- renamed step `VerifyUnspent` to `UnspentBind`

### DIFF
--- a/crates/tachyon/src/primitives/anchor.rs
+++ b/crates/tachyon/src/primitives/anchor.rs
@@ -14,9 +14,8 @@ use crate::{digest::poseidon, serialization};
 /// - [`Anchor::next_stamp`] (`Tachyon-StampFld`) absorbs one stamp's epoch and
 ///   tachygram-set commitment.
 /// - [`Anchor::next_epoch`] (`Tachyon-EpochStp`) lifts across an epoch
-///   boundary. `UnspentEpochFuse` checks the boundary against the anchors it
-///   joins; `SpendableEpochLift` and `EmptyEpochUnspentSeed` fold it from a
-///   predecessor they do not constrain to be the epoch's terminal anchor.
+///   boundary. `EndEpochUnspentSeed` is the only step that folds it, from a
+///   predecessor it does not constrain to be the epoch's terminal anchor.
 ///
 /// A block that publishes no stamp contributes no link, so the anchor is
 /// constant across a stampless span.
--- a/crates/tachyon/src/primitives/anchor.rs
+++ b/crates/tachyon/src/primitives/anchor.rs
@@ -14,7 +14,9 @@ use crate::{digest::poseidon, serialization};
 /// - [`Anchor::next_stamp`] (`Tachyon-StampFld`) absorbs one stamp's epoch and
 ///   tachygram-set commitment.
 /// - [`Anchor::next_epoch`] (`Tachyon-EpochStp`) lifts across an epoch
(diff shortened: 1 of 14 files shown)


crates/tachyon/src/stamp/proof/spendable.rs
```diff
@@ -93,69 +109,137 @@ impl Step for SpendableInit {
...
+/// `(e+1, GGM(mk, e+1), A.next_epoch(e+1))`, reading both nullifiers off a
+/// two-epoch [`NullifierHeader`] tied to the lineage by `cm`.
+///
+/// A lineage on its epoch's terminal anchor has no segment to lift over: a
+/// segment's `anchor_prev` is the anchor before its first link, and the epoch
+/// has no link left.
+///
+/// # Soundness
+///
+/// That `spendable_anchor` is epoch `e`'s terminal anchor is unconstrained
```
LucidSamuel, line 194: could we add an adversarial test analogous to `spendable_init_accepts_forged_anchor`? it should show that ticking a mid-epoch anchor succeeds in-circuit but produces an anchor outside the published sequence. that would directly pin this load-bearing soundness claim.

crates/tachyon/src/primitives/anchor.rs
```diff
@@ -9,15 +9,16 @@ use crate::{digest::poseidon, serialization};
...
 ///
-/// A Poseidon hash sequence with three domain-separated link types:
+/// A Poseidon hash sequence with two domain-separated link types:
 ///
 /// - [`Anchor::next_stamp`] (`Tachyon-StampFld`) absorbs one stamp's epoch and
 ///   tachygram-set commitment.
-/// - [`Anchor::next_empty`] (`Tachyon-EmptyBlk`) advances through one block
-///   that contains zero stamps, preserving per-height anchor uniqueness.
 /// - [`Anchor::next_epoch`] (`Tachyon-EpochStp`) lifts across an epoch
 ///   boundary; checked against a boundary chain's root by `SpendableInit`.
```
LucidSamuel, line 17: this is stale now that `SpendableInit` no longer consumes an `AnchorChain`. the current explanation should point to `UnspentEpochFuse` for checked boundary continuity and to `SpendableEpochLift`/`EmptyEpochUnspentSeed` for the computed boundary folds.

LucidSamuel (commented):
lgtm, left some nits

crates/tachyon/src/stamp/proof/spendable.rs
turbocrime: this is, essentially, testing the property that poseidon is a digest for which different preimage produces different output. i'm not sure we need tests in our repo that confirm this.

the free-witness characteristics and soundness rationale are already described in rustdoc and the book.

i think i would actually prefer to simply delete spendable_init_accepts_forged_anchor

PR #187, opened 2026-08-19
refactor: horner windows

turbocrime:
After #183 

Changes how a nullifier sequence is encoded. No steps are added or removed.

The coefficient order reverses. For members `n_0..n_{k-1}`,
`N(X) = n_0·X^{k-1} + ... + n_{k-1}`, so the oldest member takes the top degree and
the latest sits at degree 0.

- appending a member is `N ↦ X·N + n`
- concatenating two sequences is `merged = X^{k_right}·left + right`

A sequence now includes its present value. `elapsed` holds one member per covered
epoch over the inclusive range `[epoch_start, epoch_last]`, with the in-progress
nullifier at degree 0. Previously it held one member per crossed boundary, and the
present nullifier sat outside the sequence.

The sentinel coefficient is removed. It kept an empty sequence off the identity point
and pinned the sequence rank. In its place:

- an empty sequence is not representable; `NfSeqPoly::from_iter` requires a member
- the rank is pinned by the announced epoch spans and a nonzero top coefficient
- the top coefficient is threaded from `NfDerive` through `NullifierFuse`

Consumer changes:

- `UnspentBind` reads `elapsed` directly, with no `cap_member` substitution
- `UnspentBind` equates `nf_start` and `nf_last` only for a single-epoch segment
- `enforce_covering_read` no longer takes a `cap_member` argument
- the older margin now sits above the read, the cap-shifted tail below it
- scalar read members are supplied oldest-first

Renames, following the now-inclusive bound:

- `nf_end` to `nf_last`
- `epoch_end` to `epoch_last` on `ArbitraryUnspent` and `Unspent`
- `nf_seq_commit` to `nf_commit` on `NullifierDerivation`

Tests: `primitives/seq.rs` covers the append identity and the coefficient order.
Fourteen proof-step tests are added, including:

- `spend_bind_rejects_a_solved_next_over_garbage_margins`, and the
  `spendable_epoch_lift` equivalent, covering a prover who solves for the next
  nullifier by choosing the margins
- `multi_window_span_binds_once`
- `sentinel_less_empty_margin_cannot_be_witnessed`
- `spendable_epoch_lift_crosses_a_terminal_anchor_to_a_spend`


turbocrime:
folding this into #183 

this is probably superseded by #192

PR #191, opened 2026-08-21
feat: provide bundle and stamp lifting methods

turbocrime:
`Bundle<ProofStamp>::lift` is the primary consumer api for use with whole consensus data.

consumers desiring more precise control may use
- `ProofStamp::lift` accepts witness data and builds `AnchorChain`
- `ProofStamp::prove_lift` accepts `AnchorChain` pcd directly

anchor advancement is now fallible, with `AnchorError`

additionally, `ProveError` unifies some cases to `MissingPcd`

### DIFF
--- a/crates/tachyon/src/bundle/mod.rs
+++ b/crates/tachyon/src/bundle/mod.rs
@@ -291,9 +291,6 @@ pub enum LiftError {
     /// No lift is needed.
     #[display("no lift is needed")]
     NoLift,
-    /// Action digest construction failed (cv or rk was the identity point).
-    #[display("action digest failed: {_0}")]
-    ActionDigest(ActionDigestError),
     /// A provided anchor input could not advance the anchor.
     #[display("anchor step failed: {_0}")]
     AnchorError(AnchorError),
@@ -302,7 +299,7 @@ pub enum LiftError {
     AnchorContinuityFailed(ragu::Error),
     /// The stamp lift itself failed
     #[display("stamp lift failed: {_0}")]
-    LiftFailed(ragu::Error),
+    LiftFailed(stamp::ProveError),
 }
 
 /// Errors during bundle verification.
@@ -567,14 +564,11 @@ impl Bundle<ProofStamp> {
         adjuncts: &[&Bundle<dyn StampState>],
         (epoch, tachygram_sets): (EpochIndex, Vec<TachygramSetCommit>),
     ) -> Result<Self, LiftError> {
-        let own_digests = self.actions.iter().map(|&action| action.digest());
-        let other_digests = adjuncts
-            .iter()
-            .flat_map(|&adj| adj.actions.iter().map(|&action| action.digest()));
-        let action_digests = own_digests
-            .chain(other_digests)
-            .collect::<Result<Vec<ActionDigest>, ActionDigestError>>()
-            .map_err(LiftError::ActionDigest)?;
+        let descriptors: BTreeSet<action::Descriptor> = self
+            .descriptors()
+            .into_iter()
+            .chain(adjuncts.iter().flat_map(|&adj| adj.descriptors()))
+            .collect();
 
         let anchor_chain = {
             let seed_witnesses: Vec<(Anchor, EpochIndex, TachygramSetCommit)> = tachygram_sets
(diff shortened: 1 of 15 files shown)


crates/tachyon/src/bundle/mod.rs
```diff
@@ -530,6 +551,57 @@ impl Bundle<ProofStamp> {
...
+            .collect::<Result<Vec<ActionDigest>, ActionDigestError>>()
+            .map_err(LiftError::ActionDigest)?;
+
+        let anchor_chain = {
+            let anchor_seeds = anchor_inputs
+                .into_iter()
+                .scan(self.stamp.anchor, |running_anchor, tg_set| {
+                    Some(
+                        PROOF_SYSTEM
+                            .seed(rng, pool::AnchorSeed, (*running_anchor, epoch, tg_set))
```
LucidSamuel: blocking: I think this needs one guard before landing because `TachygramSetCommit` can currently be constructed or decoded as the identity point, and that value eventually reaches `Anchor::next_stamp`, which asserts that the commitment is nonidentity. 

also could we either reject identity commitments here maybe with an error variant carrying the bad commitment or make nonidentity an invariant of `TachygramSetCommit`
~ a regression test showing that lift returns an error rather than unwinding would be good too

crates/tachyon/src/bundle/mod.rs
```diff
@@ -530,6 +551,57 @@ impl Bundle<ProofStamp> {
         }
     }
 
+    /// Advance the stamp's anchor across the stamps that follow it.
```
LucidSamuel, line 545: can we spell out the caller contract here? this is probably the method most callers will find first, and there are a few important requirements that aren’t obvious from the signature.

crates/tachyon/src/bundle/mod.rs
```diff
@@ -530,6 +551,57 @@ impl Bundle<ProofStamp> {
...
+    pub fn lift<RNG: RngCore + CryptoRng>(
+        self,
+        rng: &mut RNG,
+        adjuncts: &[&Bundle<dyn StampState>],
```
LucidSamuel, line 555: since this wrapper already has the bundle and its adjuncts, would it  make sense to call `verify_coverage` before doing the proof work? think that would might help catch the missing, extra, or duplicate adjunct actions early instead of consuming the bundle and returning a stamp that only fails later verification. 

~ the lower-level `ProofStamp::lift` can still remain caller-trusting.

LucidSamuel (commented):
asides non blocking nits, I only found one thing that I think needs fixing before this lands: `Bundle::lift` can panic on an identity tachygram-set commitment even though it returns Result and i left the details inline.

crates/tachyon/src/bundle/mod.rs
turbocrime: i was deliberately avoiding this to limit scope, but it turns out the diff isn't
that bad.  so now in 3d3fd6452613f7fd02930b32cba9e1a64ee193f4 anchor methods are `Result<Anchor,AnchorError>` 

it's not useful for `LiftError` to carry the problematic commitment since it
would always be the same, but we now have `LiftError::AnchorError`

eventually we should transition set commit newtypes to `EqAffine` but i am going
to cut scope here.

tests are added.

crates/tachyon/src/bundle/mod.rs
turbocrime: updated docs

crates/tachyon/src/bundle/mod.rs
turbocrime: no. same reasoning as https://github.com/tachyon-zcash/tachyon/pull/176#issuecomment-5075820158

the caller is expected to deliberately select suitable inputs. by the time the
caller is preparing to lift, they understand the coverage situation.

callers that want to `verify_coverage` have already done so, and we should not
repeat the check.

crates/tachyon/src/bundle/mod.rs
turbocrime: 89a13f6f provides two failure conditions.

an identity commit fails with `AnchorError::NextStampZero`

an empty set `poly_with_roots(&[])` is not the identity, and fails with `AnchorError::NextStampEmpty`

crates/tachyon/src/bundle/mod.rs
turbocrime: changed the signature of this method to accept a sequence of bundles, instead of bare commits. callers handling bare commits can use the stamp methods directly.

PR #194, opened 2026-08-27
feat: expose helpers to create spend proof inputs

S1nus:
Since `PROOF_SYSTEM` is not exposed, we can either expose it somehow, or expose helpers for constructing these PCDs necessary for generating transactions.

Here is a prototype of such accessors


S1nus:
closing for now

PR #196, opened 2026-08-29
feat: bucket epoch state by quadratic residue

turbocrime:
Adds shared `QrBucket` evidence for closed-epoch exclusion and creation
membership. Exclusion stays note-independent and delegatable until
`UnspentBind`.

## Discriminants

The routing prover chooses entropy at each root; its digest under
`Tachyon-QrDiscrm` is $R_1$, carried on every QR header, and the
discriminants progress by one:

$$
R_1 = \mathsf{Poseidon}_\mathtt{Tachyon\text{-}QrDiscrm}(\mathsf{entropy}),
\qquad
R_{j+1} = R_1 + j .
$$

A split at depth $j$ classifies at $R_{j+1}$. A `true` bit means $x + R$ is
square, including zero. No QR header carries an anchor other than
`anchor_prev` and `anchor_last`, so buckets can build while the epoch is
live.

## Building an epoch's buckets

`QrSummaryIntakeInit` seeds a `QrIntake` from a `Summary`,
`QrStampIntakeSeed` from one stamp. An intake carries epoch, span, $R_1$,
profile and contents commitment.

`QrIntakeSplit` emits `QrIntakeSides` with factors $q_0$ (`non_residue`)
and $q_1$ (`residue`) of the contents $p$, checking only

$$
p(X) = q_0(X)\,q_1(X), \qquad q_0(-R) \neq 0 ,
$$

so $-R$ can occur only in `residue`.

`QrSideDescend` returns one side and checks that every root of the *other*
side takes its class at $R$; with the product, every member of the returned
class is then in the child. The returned side may carry stray members of
the other class, which only tightens the openings consumers make. The class
multiplier is linear in the bit, so no constraint branches on the witness;
the challenge absorbs the three commitments and $[R_1]G_0$.

`QrIntakeMerge` joins adjacent same-profile intakes agreeing on epoch and
$R_1$ whenever the product fits one polynomial. Sibling buckets need not
stop at the same depth.

`QrBucketSeal` alone emits a `QrBucket`, checking
$\mathsf{anchor\_prev} = \mathsf{Poseidon}_\mathtt{Tachyon\text{-}AnchorEp}(\mathsf{prev\_last}, e)$.
That `anchor_last` is the epoch's terminal anchor closes through the
consuming lineage, as a summary's span does.

## Proving a nullifier absent from an epoch

`QrUnspentInit` takes a `QrBucket` alone and witnesses a nonzero $x$, a side
$b_j$ and root $r_j$ at each of the 32 positions, a depth mask $m_j$, the
one-member sequence naming $x$, and the contents. With $s_j = x + R_1 + j$
and $c = 5$,

$$
r_j^2 = \bigl(c - (c - 1)\,b_j\bigr)\, s_j,
\qquad
b_j = 0 \implies s_j \neq 0,
$$

fixes the value's own side at every position, filing $s_j = 0$
residue-side as the split does. The mask selects the bucket's depth $d$ as a
prefix and a fold compares its sides against the value's:

$$
\sum_j m_j = d,
\qquad
2 \sum_j j\, m_j = d\,(d - 1),
\qquad
a_{j+1} = a_j + m_j\,(a_j + b_j),
\qquad
a_{32} = \mathsf{bits} .
$$

The step then binds the sequence at a challenge absorbing $[x]G_0$, opens
the bucket at $x$ for nonzero, and emits an `ArbitraryUnspent` over the
bucket's span. Two oracles. Consecutive epochs compose through
`EndEpochUnspentSeed` and `UnspentFuse`.

## Starting a spendable from a bucket

`QrSpendableInit` takes the note's creation-epoch `Unspent` and a bucket,
opens the bucket at $\mathsf{cm}$ for zero, requires the spans to match, and
emits at the segment's terminal anchor. Membership needs no profile: split
and merge add no roots outside their inputs.

## Cost

Seven steps, none above three oracles. A level costs $3n + m$ steps for $n$
intakes and $m$ merges; an epoch query is one step. `QrProfile::MAX_DEPTH`
is 32; honest depth stays below 26 at any throughput proposed. Gate costs
are estimates, since the mock measures nothing.

## Divergences from `book-alex`

| `book-alex` term | Implementation |
| --- | --- |
| QR offset $R$ | `QrDiscriminant` |
| `Root` / intermediate bucket | `QrIntake` |
| `Partition_s` | `QrIntakeSides` |
| `Tachygrams` | `QrBucket` |

1. **Reuse `Summary` (#188)** in place of counted `RootAppend` batches.
2. **Split, descend, merge** in place of paired partition; one-input splits
   are prototype scope, not a step limit.
3. **Check the side not returned**, eliminating separate check states.
4. **Prover-chosen $R_1$** with an additive progression, in place of the
   closing sentinel, so buckets build during the epoch and a query derives
   no hashes.
5. **Single-value claims**: the QR segment excludes the closing tick;
   crossings carry both epochs' values. An interface choice.
6. **Split bootstrap**: `QrSpendableInit` reuses the note-bound `Unspent`
   and tests $\mathsf{cm}$ membership directly.

## Mock-ragu gaps

`QrSideDescend` needs constraints for its boolean bit and `depth < 32`;
`QrUnspentInit` for booleanness of every side and mask bit. Challenges bind
field elements as $[x]G_0$ because mock `derive_challenge` accepts only
points.

## Tests

41 integration tests in `crates/tachyon/tests/integration/qr.rs`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -396,15 +396,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "maybe-rayon"
-version = "0.1.1"
-source = "git+https://github.com/tachyon-zcash/maybe-rayon.git?rev=5c37ee0dd448e8c1edc4529d0b8897ecaff5aad1#5c37ee0dd448e8c1edc4529d0b8897ecaff5aad1"
-dependencies = [
- "cfg-if",
- "rayon",
-]
-
-[[package]]
 name = "memchr"
 version = "2.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -493,46 +484,43 @@ checksum = "dc33ff2d4973d518d823d61aa239014831e521c75da58e3df4840d3f47749d09"
 [[package]]
 name = "ragu"
 version = "0.0.0"
-source = "git+https://github.com/tachyon-zcash/ragu?rev=f4f6315b10ed96944c40b2bbdbca0fe06af66ce0#f4f6315b10ed96944c40b2bbdbca0fe06af66ce0"
+source = "git+https://github.com/tachyon-zcash/ragu?rev=0881656dc3d9f47d3d9eaaf692b15dd8938d9bf9#0881656dc3d9f47d3d9eaaf692b15dd8938d9bf9"
 dependencies = [
  "blake2b_simd",
  "ragu_arithmetic",
- "ragu_circuits",
  "ragu_core",
  "ragu_pasta",
- "ragu_pcd",
- "ragu_primitives",
 ]
 
 [[package]]
 name = "ragu_arithmetic"
 version = "0.0.0"
-source = "git+https://github.com/tachyon-zcash/ragu?rev=f4f6315b10ed96944c40b2bbdbca0fe06af66ce0#f4f6315b10ed96944c40b2bbdbca0fe06af66ce0"
+source = "git+https://github.com/tachyon-zcash/ragu?rev=0881656dc3d9f47d3d9eaaf692b15dd8938d9bf9#0881656dc3d9f47d3d9eaaf692b15dd8938d9bf9"
 dependencies = [
  "ff",
  "group",
- "maybe-rayon",
  "ragu_macros",
  "rand",
+ "tachyon-maybe-rayon",
  "zakura-pasta-curves",
 ]
 
 [[package]]
 name = "ragu_circuits"
 version = "0.0.0"
-source = "git+https://github.com/tachyon-zcash/ragu?rev=f4f6315b10ed96944c40b2bbdbca0fe06af66ce0#f4f6315b10ed96944c40b2bbdbca0fe06af66ce0"
+source = "git+https://github.com/tachyon-zcash/ragu?rev=0881656dc3d9f47d3d9eaaf692b15dd8938d9bf9#0881656dc3d9f47d3d9eaaf692b15dd8938d9bf9"
 dependencies = [
  "blake2b_simd",
- "maybe-rayon",
  "ragu_arithmetic",
  "ragu_core",
  "ragu_primitives",
+ "tachyon-maybe-rayon",
 ]
 
 [[package]]
 name = "ragu_core"
 version = "0.0.0"
-source = "git+https://github.com/tachyon-zcash/ragu?rev=f4f6315b10ed96944c40b2bbdbca0fe06af66ce0#f4f6315b10ed96944c40b2bbdbca0fe06af66ce0"
+source = "git+https://github.com/tachyon-zcash/ragu?rev=0881656dc3d9f47d3d9eaaf692b15dd8938d9bf9#0881656dc3d9f47d3d9eaaf692b15dd8938d9bf9"
 dependencies = [
  "ragu_arithmetic",
  "ragu_macros",
@@ -542,7 +530,7 @@ dependencies = [
 [[package]]
 name = "ragu_macros"
 version = "0.0.0"
-source = "git+https://github.com/tachyon-zcash/ragu?rev=f4f6315b10ed96944c40b2bbdbca0fe06af66ce0#f4f6315b10ed96944c40b2bbdbca0fe06af66ce0"
+source = "git+https://github.com/tachyon-zcash/ragu?rev=0881656dc3d9f47d3d9eaaf692b15dd8938d9bf9#0881656dc3d9f47d3d9eaaf692b15dd8938d9bf9"
 dependencies = [
  "num-bigint",
  "num-traits",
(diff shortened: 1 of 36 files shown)


alxiong (commented):
## Agreed

👍 These are diff from my writeup that I agree with, and will update my proof tree towards.

- "Split-Descend-Merge" over a 2-input-2-output "paired partition". 
  - Mine needs to manage expected partitioned size probabilistically, and i agree that on-demand merge feels easier.
  - the `QrSideDescend (QrIntakeSides) -> QrIntake` for each side is also much nicer than my ugly multi-step partition checks: `AnchorPartition_{permuted, left_checked, checked}`
- the summarized diff point 6: "Split bootstrap". This is due to an honest mistake in my current writeup, my `SpendableReinit` is intended to be exactly as your `QrSpendableInit`, but accidentally i wrote that i'll test `nf_incl` in that step as well, which is false and unintended. `SpendableReinit` is supposed to be only about inclusion and doesn't touch `Unspent` side at all.
- I don't need to re-open `cm` in the `SpendableReinit`/`QrSpendableInit` step, just blindly test it and carry it, since `UnspentBind` already test its derivation; Also, I don't need to prove any anchor-chain ancestry from the inclusion anchor to the ending sentinel anchor. Both are my mistake and I'll follow yours.
- Anchor chain evidence is only used for the active epoch to lift stamp, never any past epochs
- Don't need a QR-tree for anchor chain evidence

## To Be Discussed

These are design decisions that I'm not fully convinced of, thus would appreciate some discussion. 

### Multi-step `QrProfile` enforcement

Currently, a whole mechanism including `QrFilter`, `QrProfileClaim` etc exist under the assumption that we cannot enforce a full `QrProfile` in a single step, thus must be enforced across steps, and partially-tested profile are accumulated in `QrFilter`. 

I'd like to challenge this. 

Take the old trick: 4 QR discriminant per Poseidon squeeze. we gets 4*7=28 `R_j` per step. Now, let me argue `QrProfile.bit` only needs to be `u32`, not even `u64`. 

Assuming 50k TPS (which is absolutely wild and multi-year long consensus engineering goal) and 2 week epoch length, all tx are 2-in-2-out, thus 8 tachygrams. We get 483 billion tachygrams per epoch, assuming 8000 tg per `QrIntake` bucket, we gets at most 60 million such buckets, and `k = log2(60 mil) < 26`. Therefore 32 depth is more than enough, 64 is an overkill.

Given this, the ideal `QrProfileAttest` step in my mind should only derive all these 26 discriminant, and spend 5 `mul` (2 for QR, 3 for NQR) per discriminant, another additional 1 `mul` per bit for boolean enforcement. We need another 26 `mul` for decomposition of the encoded profile field (i.e. all profile bits add up to `QrProfile.bits`)
Napkin calculation tells me we can enforce a full profile within a single step.

:point_right:  So a simplified `QrProfileClaim { epoch, boundary, profile, tachygram }` should work.

### Non-sentinel end anchor for `Unspent`

Currently `ArbitraryUnspent` can carry an ending anchor of any in-epoch anchor. When crossing the epoch boundary, we first get an `ArbitraryUnspent` until the `terminal_e` which is the last stamp in epoch e, then do a `UnspentFuse` with output from `EndEpochUnspentSeed`. 

I find this two-step "boundary cross" unnecessary and prefer the single `UnspentLift` model in my writeup where each `Unspent` header strictly covers a starting and ending anchors both being sentinels. (thus i we can track using a counter, starting epoch `s` and number of epoch elapsed `m`, and specify the starting `sentinel_s` and the ending `sentinel_s+m`.) Both sides are inclusive. 
Semantically, when we see `Unspent(sentinel_e, sentinel_e+1, ...)`, we interpret this as the range in `e` even though the inclusive ending is technically a special "starting point" of the next epoch. 
This doesn't contradict with the following:

>5. Keep single-value claims. Unlike UnspentLift, the QR segment excludes
the closing tick. Existing ArbitraryUnspent crossings carry both epochs'
values; separating the tick keeps QrProfileClaim single-valued. This is
an interface choice, not a QR or delegation requirement.

With full-epoch coverage `UnspentLift`, we can still keep `QrProfileClaim` single-valued. There's no ambiguity on which `nf_e` are we testing for. nor is there any need to include two `(nf_e, nf_e+1)` and give up single-valued. 

:point_right: In effect, I propose we rename `ArbitraryUnspent` to `Unspent`, remove `EndEpochUnspentSeed` step as it's no longer needed. `QrBucket.anchor_last` should be the sentinel anchor instead of "terminal anchor"; so that mid-epoch/non-sentinel `anchor_last` never happens in our proof tree.
(btw, as a consequence, `SpendableHeader.present_nf` can be removed because it's used as boundary token in arbitrary-span model) 

I grill AI to make sure I didn't miss anything, and it agrees with me:

>  the crossing's three jobs are D7's live-value naming (your counters absorb it), in-circuit tick-folding for segments ending at stamped anchors (only sub-epoch producers need it — the QR path's seal already folds the tick, making the crossing a redundant second fold there), and the single-valued claim (already shown conditional). 
> **Your UnspentLift achieves everything the crossing does within the quantized model — and does it cheaper,** because the crossing is per-note work (it witnesses your lineage's own nullifier pair) while the sentinel-inclusive tick fold amortizes into the shared bucket seal, once per epoch for all wallets.

turbocrime:
the `UnspentLift` is omitted to close scope and limit complexity of the initial implementation

turbocrime:
during work on this i eliminated cap logic since u64 is the natural `Fp::from` input. i'm pushing updates soon that introduce a 32-bit cap as requested.

turbocrime:
your depth bound of 26 works for historical evidence, and if buckets are only
ever built after the epoch closes, 32 is fine.

however i would like a system in which buckets are assembled while the epoch is
in-flight, so they can be ready almost immediately when the epoch closes. to
achieve this we may make `R` a consensus value, or we may make `R` an arbitrary
prover choice. i worked on this today and produced an implementation that uses
an arbitrary discriminant.

however, in-flight evidence is exposed to grinding risk: sync services do a lot
of work across the epoch to produce evidence, and any structural limits (bucket
size and depth) could be targeted by someone who knows the sync service's
selected discriminant. a griefer positioned as an aggregator may be able to land
these tachygrams on chain by producing large aggregates attractive to miners.

up to the depth limit, this only costs the sync service prover only a few more
steps per split. however, *at* the depth limit, there are no more splits, and
the evidence simply can't be completed. the targeted sync service would have to
discard all existing epoch work and rebuild with a new discriminant. late in the
epoch, this could be very costly.

specifically this attack requires landing more than 8192 tachygrams sharing one
full-depth profile. finding each one costs $2^d$ hashes on average, so the full
set is $8192 \cdot 2^d$ hashes, and then about 1 ZEC in fees to make the
transactions.

at $d = 32$ that's $2^{45}$ digests (days on gpu). at $d = 64$ it's $2^{77}$ and
much more difficult.

in-flight evidence, built by more than one party, probably needs 64.

i guess this means the branch i'm pushing right now is kind of a mess, since i
reintroduced the 32-bit cap, but the implementation is designed for in-flight
building.

turbocrime:
we might simply accept the situation, and ask sync services to not publish their chosen discriminant until the epoch is closed.

if we revert to the 64-bit cap, we could fix the discriminant as a consensus value available at epoch start, to permit broad sharing of evidence.

turbocrime:
i'm also concerned that the total size of epoch evidence is going to be too large. i will send you some cost estimates.

turbocrime:
i didn't align on the boundary-anchor question, i would like to defer that to later iteration

alxiong:
all the "agreed changes" summarized in https://github.com/tachyon-zcash/tachyon/pull/196#pullrequestreview-5141009605 has been reflected in [20f0e43](https://github.com/tachyon-zcash/tachyon/pull/129/commits/20f0e4344eee6bc497f165b7f6be78419f04605b) and [90b1c1a](https://github.com/tachyon-zcash/tachyon/pull/129/commits/90b1c1a981b1d7317e5ac414bc6cccb5568ac003) of my writeup PR.

Open discussion:
first suggestion on removing multi-step QrFilter is removed in [5d08243](https://github.com/tachyon-zcash/tachyon/pull/196/commits/5d08243c24449bdab9c0f170601aac7a4f59f845), 
second suggestion will be addressed in future PRs.

turbocrime:
grinding is no longer a concern in 465a5f4854b0d28e23c398f1752c2ca5870f6dab which aligned on historic-only epoch evidence, discriminated by the transition anchor.

PR #198, opened 2026-08-31
note: rename Positive to NonNegative

alxiong:
While #182 already allows zero-value for note. The name stayed as `Positive`. 

This chore PR did the following renaming:
- `Postive` -> `NonNegative`
- `Negative` -> `NonPositive`

no logic changes.


turbocrime:
i think it is fine to keep the names Positive and Negative

alxiong:
I don't mind calling it `NoteValue` as in [orchard](https://github.com/zcash/orchard/blob/be4f659467338a86e93bd3d9ae72a696bebc924b/src/value.rs#L100). 

`Positive` is simply wrong.

PR #200, opened 2026-09-01
Tighten clippy lint

alxiong:
Some code quality chores as I scan through the repo.

For consensus/critical/crypto libraries, I think it's better to set `arithmetic_side_effects = "deny"` (with all upstream field arithmetic as whitelist exceptions). 
This nudge us to correctly use unambiguous `saturating_*()` or `.wrapping_*()` or justify with reason or avoid them completely.

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -28,7 +28,7 @@ jobs:
   notify-website:
     name: Notify website to deploy
     needs: build
-    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main' && github.repository == 'tachyon-zcash/tachyon'
     runs-on: ubuntu-latest
     steps:
       - name: Trigger repository_dispatch in website
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -144,6 +144,7 @@ rc_mutex = "forbid"
 
 # -- deny: restrictions (may need localized #[expect]) -----------------------
 absolute_paths = "deny"
+arithmetic_side_effects = "deny"
 assertions_on_result_states = "deny"
 big_endian_bytes = "deny"
 cfg_not_test = "deny"
@@ -207,7 +208,6 @@ verbose_file_reads = "deny"
 
 # deny later
 arbitrary_source_item_ordering = "allow"
-arithmetic_side_effects = "allow"
 exhaustive_enums = "allow"
 exhaustive_structs = "allow"
 missing_errors_doc = "allow"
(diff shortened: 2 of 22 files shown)


LucidSamuel:
interesting, noticed that one boundary case appears to remain, `EPOCH_MAX` is `BLOCK_MAX/EPOCH_SIZE`, not `u32::MAX`, so `EpochIndex::next()` only checks the representation limit rather than the protocol limit.

so would it make sense for `last_block()` to calculate from the current epoch’s first block instead? 
@alxiong

alxiong:
that's a good point. fixed both issues. @LucidSamuel

turbocrime (changes requested):
the new `EPOCH_MAX` bound on `EpochIndex::next` panics some callers handling exclusive range endpoints.

for example`witness::spend_bind` near the last epoch may compute an empty sequence positioned at `epoch.next().next()` just beyond the epoch space. here, we could conditionally use `NfSeqPoly::default()` (or a quotient operation from #201) to avoid attempting to compute the start index of the empty sequence.

i think we should avoid panic by either using a saturating add or making the method falliable.

then `next()` callers need auditing, and we need tests that exercise edge cases near the end of the epoch space.

alxiong:
Addressed in 20a8560 (on top of the rebase onto main). Decisions, in order of your points:

`next()` stays panicking, `checked_next()` added. I went with fallible rather than saturating: saturating would silently alias two distinct epochs at `EPOCH_MAX,` which corrupts sequence positions instead of failing. But rather than making every caller `unwrap` an `Option,` the two methods now express the two semantics — `next()` for "a successor must exist here" (panic = caller bug), `checked_next()` where the final epoch is ordinary input.

Exclusive endpoints in witness prep no longer step past the end. Your `NfSeqPoly::default()` suggestion works directly: encode(∅) folds from constant 1, so the empty run is the multiplicative identity and needs no position at all. There's now a `complement_tail(span_last, tail)` helper in `witness.rs` that returns the identity for an empty tail and only computes `span_last.next()` when the tail is non-empty (in which case the window provably extends past `span_last,` so the successor exists). Used by `unspent_bind`, `spendable_init`, `spend_bind`, and `summary_spendable_init`; `unspent_bind`'s `hi` is also computed without `next()` now.

turbocrime:
pulled out epoch work into #210, let's merge that first so this becomes actually focused on lint config and toolchain

provisional rebase/port of this branch https://github.com/turbocrime/tachyon/tree/tighten-clippy-lint

alxiong:
can you submit your branch as a PR instead? i'm closing this.
@turbocrime

PR #201, opened 2026-09-01
Set representation and lazy poly realization

alxiong:
Addressing my own comments in: https://github.com/tachyon-zcash/tachyon/pull/183#pullrequestreview-5066315864:

> For the encoded polynomial for an indexed multiset, I believe the right approach is to delay "realization into ragu::Polynomial as late as .commit() time. We should instead just keep the HashSet<(usize, Fp)> as long as we could because majority of the operation are much cheaper by operating on this.
>
> The most important reason is to avoid polynomial division when computing the quotient witness, if we keep it as a HashSet, then the q(X) computation is as easy as set subtraction, not even field ops involved.
> 
> multiplication is also cheap set union. evaluation is what you already have as direct_eval, but i'd just make it a method as IndexedMultisetPoly::eval() which computes in streaming fashion. this is definitely faster than .eval() on a realized ragu::Poly because we can massively parallelize the evaluation.

This this PR:

-  refactor: set-based indexed multiset with lazy realization — IndexedMultiset keeps a `BTreeMap<(u64, Fp), NonZero<u32>>` count-map as the canonical form (answering turbocrime's multiplicity objection), memoizes both the realized `Polynomial` and its commitment (killing `NfDerive`'s triple realize+MSM), adds `union/quotient` (count merge / count subtraction — `None` doubles as the subset check)
- removed `poly_mul()` helper func, as we agree in https://github.com/tachyon-zcash/tachyon/pull/183#discussion_r3894611349
-  chore: introduce `multicore` feature (same as Ragu)
-  perf: parallel indexed multiset evaluation — `eval()` maps member factors through `into_par_iter()/.product()`


crates/tachyon/src/collections/indexed_multiset.rs
```diff
@@ -103,29 +119,229 @@ fn direct_eval_single(idx: u64, m: Fp, x: Fp) -> Fp {
...
+        let members: Vec<((u64, Fp), NonZero<u32>)> = self
+            .members
+            .iter()
+            .map(|(&key, &count)| (key, count))
+            .collect();
+        members
+            .into_par_iter()
+            .map(|((idx, m), count)| {
+                let factor = direct_eval_single(idx, m, x);
+                (1..count.get()).fold(factor, |subtotal, _| subtotal * factor)
```
LucidSamuel: this is linear in the multiplicity as `factor.pow([count.get().into()])` would do it in log steps. it barely matters since counts are ~1, but insert tolerates up to u32::MAX so eval may as well not scale with it.
- non blocking btw

LucidSamuel (approved):
LGTM, left one non blocking nit inline

crates/tachyon/src/collections/indexed_multiset.rs
alxiong: good idea, i changed to:

```rust
                if count.get() == 1 {
                    factor
                } else {
                    factor.pow([NonZero::<u64>::from(count).get()])
                }
```
and then fixup to the last commit, force-pushed.

turbocrime (changes requested):
at minimum this needs rebase, because it is broken when applied to the tip of main.

imo we should avoid adding the complexity of feature-conditional multithreading optimizations for now. let's defer things like that until we can make a whole-project optimization pass. fwiw i actually observe performance regression here.

turbocrime:
if we introduce types to manage these things, i would like to see them applied uniformly (multiset and indexed multiset). edit: fold in #203 

as a goal, any implementation should reduce complexity.

alxiong:
rebased, `multicore` commits reverted, and cherry-pick #203 over. @turbocrime

turbocrime (changes requested):
all locations that need a complement to some subsequence produce it by construction. nothing calls the quotient method.

so as it stands this branch adds a complex newtype that nothing actually takes advantage of.

the memoized commitment is the most useful part, but i'm not sure that we actually should/need to be providing that feature.

alxiong:
>all locations that need a complement to some subsequence produce it by construction. nothing calls the quotient method.

this is because we will need to update our stamp witness generation to avoid using polynomial division to come up with the quotient. and I haven't done that.
any polynomial multiplication and division, which we already do in our proof tree to build up the QrBucket benefit from this. 

I was just trying to keep the scope small.

turbocrime:
ok. stack with another PR demonstrating consumer refactors and i will re-review

alxiong:
sounds good, the net effect can be seen in the last commit: [f2d3e18](https://github.com/tachyon-zcash/tachyon/pull/201/commits/f2d3e1812e1a8c9e85422ffa58f50c004e98b4cd)

where you would still use the `*` and `/` on these linear-factorized polynomials, but underneath them are super-efficient set operations involving no actual field ops or expensive inversion (which is carry out GCD). 

there's just one exception:
> `collections::qr::decomposition`'s `divide_by_root.` That divides $u^2 − c(X+R)$ by the sibling's vanishing polynomial; the dividend isn't a multiset product, so its quotient has roots we don't know and set subtraction can't produce it. It's the only polynomial division left in production code.

I also introduced `trait FactoredPoly` and a macro `impl_factored_poly!` that reduce code duplication in the three instantiation of root-set-based poly.

turbocrime:
spent some time looking at this today. rebased and tried to simplify. but as i said during meeting, it feels like this kind of work premature and of unclear benefit given the present reliance upon the mock.

i don't know how to reason about it now, so i'd prefer to just avoid the added complexity. we should revisit soon.

PR #202, opened 2026-09-01
Introduce Anchor Types

alxiong:
Part of spec alignment, currently the only difference is whether `Anchor` is derived from `.next_stamp()` or `.next_epoch()` where the latter is the sentinel anchor. 

In the PR, we introduce a sealed, zero-sized type  `trait AnchorKind` and modify `struct Anchor<AnchorKind>` to make it type-level differentiable the type of anchor it is. 
There are three types `Stamped` (the normal anchor), `Sentinel` the beginning anchor of the next epoch, `Erased` is the type-erased variant that doesn't require any precedessor chain to instantiate, just an arbitrary value. 

At this point, this `AnchorKind` has no material changes to circuit logic or benefit yet. But this prepares us for the epoch-wide `AnchorChain` and `Tachygrams` evidence mentioned in #129. Because those headers require explicit sentinel anchors in their header, and not just any arbitrary ones.

### DIFF
--- a/crates/tachyon/src/stamp/proof/spendable.rs
+++ b/crates/tachyon/src/stamp/proof/spendable.rs
@@ -163,7 +163,7 @@ impl Step for SpendableInit {
             .next_stamp(creation_epoch, &creation_commit)
             .map_err(|_e| ragu::Error::InvalidWitness("invalid anchor step".into()))?;
 
-        Ok(((cm, (creation_epoch, present_nf), post_cm_anchor), ()))
+        Ok(((cm, (creation_epoch, present_nf), post_cm_anchor.erase()), ()))
     }
 }
 
--- a/crates/tachyon/src/bundle/mod.rs
+++ b/crates/tachyon/src/bundle/mod.rs
@@ -565,7 +565,7 @@ impl Bundle<ProofStamp> {
                     .map_err(LiftError::AnchorError);
 
                 Some(next_anchor.map(|anchor| {
-                    *scanning_anchor = anchor;
+                    *scanning_anchor = anchor.erase();
                     (prev_anchor, epoch, tachygram_set)
                 }))
             })
--- a/crates/tachyon/src/primitives/anchor.rs
+++ b/crates/tachyon/src/primitives/anchor.rs
@@ -1,11 +1,16 @@
+use core::marker::PhantomData;
+
 use corez::io::{self, Read, Write};
-use derive_more::{Debug, Display, Eq as TotalEq, Error, From, Into, PartialEq};
+use derive_more::{Debug, Display, Eq as TotalEq, Error, PartialEq};
 use ff::Field as _;
 use group::{Curve as _, Group as _};
 use lazy_static::lazy_static;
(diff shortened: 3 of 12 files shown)


crates/tachyon/src/stamp/proof/spendable.rs
```diff
@@ -163,7 +163,7 @@ impl Step for SpendableInit {
             .next_stamp(creation_epoch, &creation_commit)
             .map_err(|_e| ragu::Error::InvalidWitness("invalid anchor step".into()))?;
 
-        Ok(((cm, (creation_epoch, present_nf), post_cm_anchor), ()))
+        Ok(((cm, (creation_epoch, present_nf), post_cm_anchor.erase()), ()))
```
LucidSamuel, line 166: cargo fmt --all fails on this line due to the added .erase() pushing it over length, probably wrap Ok((...))

LucidSamuel (approved):
LGTM.
nice that the inner Fp went private along the way.

turbocrime:
i think we should not do this.

in all locations where this information is meaningful, we are handling an erased type. the developer must always establish the relationship by computation or constraint.

in locations where this information is not meaningful, this change only introduces additional ceremony, and is potentially misleading.

alxiong:
> in all locations where this information is meaningful, we are handling an erased type. the developer must always establish the relationship by computation or constraint.

is it? 

here, there's no type-based protection to ensure or check if `boundary` is a sentinel. 

```rust
impl QrDiscriminant {
    pub fn of(boundary: Anchor) -> Self {
        Self(poseidon::qr_discriminant(Fp::from(boundary)))
    }
}
```

here, wouldn't it be nice to be semantically clear that this step is expecting a particularly kind of anchor? 

```rust
impl Step for QrBucketSeal {
    /// `(prev_last)`, the terminal anchor of the preceding epoch.
    type Witness<'source> = (Anchor,);
    type Output = QrBucket;
}

impl Header for QrBucket {
    /// `(epoch, anchor_prev, anchor_last, boundary, profile, discriminant,
    /// contents)`.
    type Data = (
        EpochIndex,
        Anchor,
        Anchor,
        Anchor,
        QrProfile,
        QrDiscriminant,
        TachygramSetCommit,
    );
}
```

In my mind, the following read clearer in intent and less error-prone (for implementation of the `QrBucketSeal` step as any implementor would know he has to convert `Anchor<Erased>` to `Anchor<Sentinel>`:

```rust
impl Header for QrBucket {
    /// `(epoch, anchor_prev, anchor_last, boundary, profile, discriminant,
    /// contents)`.
    type Data = (
        EpochIndex,
        Anchor<Sentinel>,
        Anchor<Sentinel>,
        Anchor<Sentinel>,
        QrProfile,
        QrDiscriminant,
        TachygramSetCommit,
    );
}
```

Sure, we eventually need to extract out the "erased `Fp`" , but the point of making it an enum-kinded is to avoid accidentally missing those checks. I thought it's a protection against AI slop at least.

but if you feel strongly that this is purely ceremonious, then I will yield. it's just a matter of taste, not correctness, so it's fine.

PR #203, opened 2026-09-02
Lazy multiset for tachygrams

alxiong:
We use the same `Multiset` representation and lazy polynomial realization for `TachygramSetPoly`; the same idea as in #201. This should improve efficiency for most operations.


LucidSamuel (approved):
LGTM, nice pow fast-path on eval.
also once the 2 clippy doc-paragraph errors  fixed (sets.rs:60/68) fixing here lets #204 inherit on rebase.

turbocrime:
fold this into #201

PR #204, opened 2026-09-02
refactor: action proofs witness multiset polys

alxiong:
This PR basically enforces the correctness of `acc^tg` in the `OutputStamp`, `SpendStamp` and `MergeStamp` steps. 

To achieve that, we introduce a collective:

```rust
/// The committed operands of one shared-challenge session: opens operands at
/// caller-chosen points, recording one claim per opening.
pub(crate) struct PolyQueries<'ctx, 'step, 'poly, const N: usize> {
    ctx: &'ctx mut StepCtx<'step>,
    operands: [&'poly Polynomial; N],
    commits: [Eq; N],
}

pub(crate) fn with_shared_challenge<'poly, const N: usize>(
    ctx: &mut StepCtx<'_>,
    operands: [&'poly Polynomial; N],
    checks: impl FnOnce(Fp, PolyQueries<'_, '_, 'poly, N>) -> Result<()>,
) -> Result<()> {}
```

that can be used as:

```rust
        with_shared_challenge(
            ctx,
            [action_set.as_ref(), tachygram_set.as_ref()],
            |z, mut queries| {
                let [action_at_z, tachygrams_at_z] = queries.open_all(z)?;
                enforce_zero(
                    action_at_z - (z - Fp::from(action_digest)),
                    "OutputStamp: action set does not commit to the action",
                )?;
                enforce_zero(
                    tachygrams_at_z - (z - Fp::from(cm)) * (z - Fp::from(pad)),
                    "OutputStamp: tachygram set does not commit to the bound pair",
                )
            },
        )?;
```

This example should epitomize the closure-based API design I intended throughout this PR.

### DIFF
--- a/crates/tachyon/src/collections/multiset.rs
+++ b/crates/tachyon/src/collections/multiset.rs
@@ -1,12 +1,340 @@
+//! # Multiset
+//!
+//! A multiset of members encoded as the roots of a monic polynomial, with
+//! multiplicity:
+//!
+//! $$
+//!   S(X) = \prod_{m \in S}{(X - m)^{\mathsf{count}(m)}}
+//! $$
+//!
+//! ## Set form
+//!
+//! Unique factorization means the member multiset *is* the polynomial, so
+//! [`Multiset`] keeps the members as its canonical representation -- a
+//! [`BTreeMap`] count-map -- and defers realization into coefficient form
+//! until [committed](Multiset::commit). Operating on the set form is much
+//! cheaper: union is a count merge rather than a coefficient convolution, the
+//! quotient witness is count subtraction rather than polynomial division, and
+//! evaluation streams over members without materializing coefficients.
+//!
+//! The stamp sets built on this type are non-repeating in practice (a stamp's
+//! tachygrams are unique on the wire), but the representation stays a general
+//! multiset: a repeated member contributes a repeated factor, which is what
+//! lets a commitment to $(X-m)^2$ stay distinguishable from one to $(X-m)$.
+
 extern crate alloc;
 
-use alloc::vec::Vec;
+use alloc::{
+    collections::{BTreeMap, btree_map::Entry},
+    vec::Vec,
+};
+use core::{cell::OnceCell, cmp::Eq as TotalEq, num::NonZero};
 
-use pasta_curves::Fp;
+use pasta_curves::{Eq, Fp};
 use ragu::{Polynomial, poly_with_roots};
 
-/// Encode the provided members as roots.
-pub(crate) fn encode(members: impl IntoIterator<Item = Fp>) -> Polynomial {
-    let roots: Vec<Fp> = members.into_iter().collect();
-    Polynomial::from_coeffs(poly_with_roots(&roots))
+/// A multiset of members held as polynomial roots, kept in set form.
+///
+/// The encoded [`Polynomial`] is realized lazily -- at first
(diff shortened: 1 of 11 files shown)


crates/tachyon/src/collections/multiset.rs
```diff
@@ -1,12 +1,340 @@
...
+
+    /// Evaluate the encoded polynomial at `x` by streaming over the members,
+    /// without realizing the coefficients.
+    #[must_use]
+    pub(crate) fn eval(&self, x: Fp) -> Fp {
+        self.members
+            .iter()
+            .map(|(&member, &count)| {
+                let factor = x - member;
+                (1..count.get()).fold(factor, |subtotal, _| subtotal * factor)
```
LucidSamuel: this _eval_ is serial while #201 parallelizes; whichever of the two lands second should reconcile

LucidSamuel (approved):
LGTM (once clippy passes)

crates/tachyon/src/collections/multiset.rs
alxiong: yeah, the parallel relies on the `multicore` feature and adding `maybe_rayon` dependencies, I'd wait for 201 to land first instead of redoing the work here.

turbocrime (changes requested):
either this needs to be an independent changeset from #201 and #203 or requested changes on those need resolution

alxiong:
>either this needs to be an independent changeset from https://github.com/tachyon-zcash/tachyon/pull/201 and https://github.com/tachyon-zcash/tachyon/pull/203

yes, this is an honest mistake, was basing off the 201 PR branch instead of main.
now I have removed unrelated commits and rebased onto main, also update some poly query usage in the QR steps that you recently land on main.
@turbocrime

crates/tachyon/src/relations/enforce.rs
```diff
@@ -1,85 +1,132 @@
...
+//! operands to open, at which points (the challenge, a statement point,
+//! several points on one operand), and which algebraic identities to check
+//! over the opened evals. Each [`PolyQueries::open`] records the opening
+//! claim `(commitment, point, eval)` and returns that same eval, so the
+//! value checked is the value opened, by construction. The relation only
+//! *records* claims; actually verifying them is the proof system's job, not
+//! done here.
+//!
+//! The challenge is shared deliberately. In practice the proof system grants
+//! a step one evaluation challenge, derived only once the whole transcript
```
turbocrime: i'm not sure this is true?

turbocrime (changes requested):
`with_shared_challenge` and related code seems like it should be ragu API rather than something we implement in tachyon.

in any case, i think the abstractions we build on top of the provisional mock api should be as few as possible. unfortunately we already have some things that cross that line. i don't want to add more.

crates/tachyon/src/relations/enforce.rs
alxiong: you can't just squeeze as many challenge as you want from the proof system, Ragu itself is just a few-round IOP, we are most likely gonna piggyback on 1 or 2 prover message to get the next verifier challenge in my  mind.

alxiong:
> with_shared_challenge and related code seems like it should be ragu API rather than something we implement in tachyon.

That's fair. I'll remove these part of the changes. But ultimately, I think it will look like this.

I'll try to push a similar mock api in Ragu instead.

update: i force-pushed and dropped these `share_challenge` api.

turbocrime (approved):
rebased and edited

PR #215, opened 2026-09-19
Rename some steps and header fields

turbocrime:
not necessarily complete or final

| old header | new header | old shape | new shape |
| --- | --- | --- | --- |
| `NfMasterHeader` | `NoteMaster` | `(cm, mk)` | `(cm, mk)` |
| `NullifierDerivation` | `NoteNullifiers` | `(cm, epoch_start, nf_commit, epoch_last)` | `(cm, epoch_first, nf_commit, epoch_last)` |
| `AnchorChain` | `AnchorChain` | `(start, end)` | `(anchor_first, anchor_last)` |
| `ArbitraryUnspent` | `ArbitraryUnspent` | `(anchor_prev, (epoch_start, nf_start), elapsed, (epoch_last, nf_last), anchor_last)` | `(anchor_prev, (epoch_first, nf_first), elapsed, (epoch_last, nf_last), anchor_last)` |
| `Unspent` | `NoteUnspent` | `(cm, anchor_prev, (epoch_start, nf_start), (epoch_last, nf_last), anchor_last)` | `(cm, anchor_prev, (epoch_first, nf_first), (epoch_last, nf_last), anchor_last)` |
| `SpendableHeader` | `NoteSpendable` | `(cm, (epoch, present_nf), anchor)` | `(cm, (epoch_current, nf_current), anchor)` |
| `SpendHeader` | `SpendHeader` | `(cm, present_nf, nf_next, anchor)` | `(cm, nf_current, nf_next, anchor)` |
| `StampHeader` | `Stamp` | `(action_commit, stamp_tg_commit, anchor)` | `(action_commit, stamp_tg_commit, anchor)` |
| `QrIntakeSides` | `QrIntakeSides` | `(epoch, anchor_prev, anchor_last, discriminant, profile, residue, non_residue)` | `(epoch, anchor_prev, anchor_last, discriminant, profile, non_residue, residue)` |

| old step | new step | old witness | new witness |
| --- | --- | --- | --- |
| `NfMasterSeed` | `NoteSeed` | `(note, pak)` | `(note, pak)` |
| `NfDerive` | `NullifierDerive` | `(epoch_start, seq)` | `(epoch_first, seq)` |
| `AnchorSeed` | `AnchorSeed` | `(start, epoch, stamp_commit)` | `(anchor_first, epoch, stamp_commit)` |
[...]

PR #216, opened 2026-09-19
refactor: prover-selected QR discriminant, boundary-to-boundary anchoring

turbocrime:
after #215

The QR discriminant is now prover-chosen and is not required to be the epoch's
closing transition anchor.

`QrUnspentInit` and `QrSpendableInit` now both handle an additional nullifier
and emit an `ArbitraryUnspent` or `NoteSpendable` anchored to the epoch's
closing transition anchor.


crates/tachyon/src/primitives/qr.rs
```diff
@@ -7,26 +7,19 @@ use ragu_arithmetic::Cycle as _;
 use ragu_circuits::polynomials::{ProductionRank, sparse::Polynomial};
 use ragu_pasta::Pasta;
 
-use super::Anchor;
 use crate::collections::qr;
 
-/// An epoch's first discriminant $R_1$.
+/// A routing network's first discriminant $R_1$.
```
alxiong: may I suggest we keep it zero-based? starting with $R_0$ and $R_j = R_0 + j$?

crates/tachyon/src/stamp/proof/qr.rs
```diff
@@ -527,11 +526,18 @@ impl Header for QrBucket {
 /// the segment binds it. Epoch zero's entry anchor, [`Anchor::default`], is
 /// this rule at an `anchor_prev_prev` of zero.
 ///
-/// Every split in the intake's history classified at $R_1 + \mathsf{depth}$
-/// read off the header, so pinning `discriminant` here pins every
-/// discriminant the routing used to the extent the bucket carries. That
-/// `anchor_last` is the epoch's *terminal anchor*, its last anchor, is
-/// likewise a claim about what was published, and this step does not check it.
+/// `discriminant` is not checked here. It is a prover-chosen routing base,
```
alxiong: reservation for "routing base", because routing has connotation to networking, so routing base sounds like the starting bucket/place (which might confuse with root bucket or so-called intakes)

I'd say either directly use $R_0$ or "first discriminant.

crates/tachyon/tests/integration/fixtures.rs
```diff
@@ -598,13 +598,9 @@ pub(crate) fn seal_qr_intake<RNG: CryptoRng>(
...
-/// The discriminant an epoch's buckets carry: the epoch link of its terminal
-/// anchor `terminal` into the next epoch.
-pub(crate) fn qr_discriminant_of(pool: &PoolSim, terminal: Anchor) -> QrDiscriminant {
-    terminal
-        .next_epoch(pool.epoch_at(terminal).next().unwrap())
-        .expect("epoch after the terminal is nonzero")
-        .into()
+/// Sample a routing base, as a builder does when it opens a network.
+pub(crate) fn qr_discriminant<RNG: CryptoRng>(rng: &mut RNG) -> QrDiscriminant {
+    QrDiscriminant::from(Fp::random(rng))
```
alxiong: remove this wrapper func? don't seem to be beneficial

crates/tachyon/src/primitives/qr.rs
```diff
@@ -7,26 +7,19 @@ use ragu_arithmetic::Cycle as _;
...
 ///
-/// The epoch-link image of the bucket's `anchor_last` into `epoch + 1`,
-/// pinned at `QrBucketSeal`. It coincides with the entry anchor of
-/// `epoch + 1` only if `anchor_last` is the epoch's terminal anchor, which the
-/// seal does not check; at `EPOCH_MAX` it absorbs an index that is no epoch.
+/// Prover-chosen and sampled privately, so a network can be routed while its
+/// epoch is still in flight. It is threaded unchanged from the root intake and
```
alxiong: nit: should add clarification somewhere in the rustdoc that:

""its own $R_0$, and its routing headers, unpublished until the epoch closes."

crates/tachyon/src/stamp/proof/qr.rs
```diff
@@ -630,9 +648,10 @@ impl Step for QrUnspentInit {
...
     type Output = ArbitraryUnspent;
     type Right = ();
-    /// `(value, classes, mask, sequence, contents)`.
+    /// `(value, nf_next, classes, mask, sequence, contents)`.
     type Witness<'source> = (
         Tachygram,
+        Nullifier,
         [QrClassRoot; QrProfile::MAX_DEPTH],
         [bool; QrProfile::MAX_DEPTH],
         NfSeqPoly,
```
alxiong: while this commit indeed achieve boundary-to-boundary and one step closer to what I depict as one `UnspentLift` per epoch, the concrete detail is slightly diff from what i have in mind.

in my mind, each `Unspent` covers a `[sntl_e, sntl_e+1)` (half-open convention) and only one nullifier `nf_e` is tested, 

each lift tests `nf_e,` multiplies in `F_{e,nf_e}`, advances to `sntl_{e+1}`, next lift derives `nf_{e+1}` itself, adjacent ranges are disjoint without any overlapping anchor. 

would that obviate the need to carry `nf_next`?
the Unspent header in my head is only :
` (epoch_start, epoch_end, sntl_start,, sntl_end, Com(g) )`

PR #217, opened 2026-09-19
refactor: action steps

turbocrime:
`NoteMaster` gains note fields.

`SpendStamp` becomes `SpendAction`, taking `NoteMaster` as its right input 

`OutputBind` and `OutputStamp` become `OutputAction`

todo: carry more nullifiers on spendable, to eliminate SpendBind?


alxiong (commented):
I agree with the merging of output steps.

More comments:

- naming: as I said in my wirteup, the rule i was following is:

> Naming: steps use "noun + verb" with verbs like "seed, fuse, lift, merge" and the noun is usually a header name, whereas headers use "adj + noun" like "Spendable, Unspent, VerifiedUnspent".

So I think `OutputStamp` or `OutputSeed` is better than `OutputAction`? we use "action" to describe the overall "spend/output action", i'm little hesitant to use it a step, especially when it's not a verb.

- adding `Note` into `NoteMaster` feels well-motivated to avoid recomputing commitment on spend side. but it breaks the abstraction where `mk` is `NoteMaster` and meant for `mk`, I'd just introduce a new header name like `NoteSecret` for the `(mk, cm, note)` combo. 
   Once we decide, I'll have to update my writeup around this, but let's discuss.

